PR #21, opened 2022-05-24
Fix/correct simulate for multi qubit gates

npoirie1:
Previously, simulating multi-qubit gates only worked if the control qubit was qubit 1 and if the target qubit was qubit 2. This pull request fixes this problem.

### DIFF
--- a/src/core/quantum_circuit.jl
+++ b/src/core/quantum_circuit.jl
@@ -239,7 +239,8 @@ function get_embed_operator(gate::Gate, system::MultiBodySystem)
     else
         gate_to_operator =
             Dict("cz"=>get_embed_controlled_gate_operator(sigma_z(), gate, system),
-            "cx"=>get_embed_controlled_gate_operator(sigma_x(), gate, system))
+            "cx"=>get_embed_controlled_gate_operator(sigma_x(), gate, system),
+            "iswap"=>get_swap(gate, system))
         return gate_to_operator[gate.instruction_symbol]
     end
 end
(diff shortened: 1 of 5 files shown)


src/core/quantum_circuit.jl
```diff
@@ -212,22 +223,99 @@ function simulate(circuit::QuantumCircuit)
...
     return ψ
 end
 
+function get_embed_operator(gate::Gate, system::MultiBodySystem)
+    if length(gate.target) == 1
+        return get_embed_operator(gate.operator, gate.target[1], system)
+    else
+        gate_to_operator =
+            Dict("cz"=>get_embed_controlled_gate_operator(sigma_z(), gate, system),
+            "cx"=>get_embed_controlled_gate_operator(sigma_x(), gate, system),
```
alireza-najafi, line 242: Oh oh!  Here you have hard coded "cz",  "cx", and "swap". This is not a good idea. We need to have a more generic approach.

src/core/quantum_circuit.jl
```diff
@@ -212,22 +223,99 @@ function simulate(circuit::QuantumCircuit)
...
+            "iswap"=>get_swap(gate, system))
+        return gate_to_operator[gate.instruction_symbol]
+    end
+end
+
+function get_embed_controlled_gate_operator(controlled_operator::Operator, gate::Gate,
+    system::MultiBodySystem)
```
alireza-najafi, line 249: This function only works with two-qubit gates. What if we have a three qubit gate such as Toffoli gate? We again need a more generic approach.

alireza-najafi (commented):
Nick, 
 Please see my comments below. The implementation of embedding multi qubit operators is not generic enough. Please see my comments for more details.

PR #31, opened 2022-08-09
Feat/add randomized benchmarking

### LINKED ISSUE #29: Add randomized benchmarking

npoirie1:
This pull request includes Clifford operators, randomized benchmarking, and new quantum circuit manipulation tools. Before merging the pull request, the get_num_qubits function in pull request #26 should be added to get_clifford_operator and the quantum gate parameters from pull request #24 should be added to the new quantum_gate functions.

Fixes #29


src/benchmarking/randomized_benchmarking.jl
```diff
@@ -0,0 +1,342 @@
...
+The models for the fit are described in [`RandomizedBenchmarkingFitResults`](@ref). The
+fit can have `model_order` 0, 1, or `nothing`. If `nothing` is specified, no fit will be
+generated. Optionally, a dictionary of initial fitting parameters can be passed to the
+constructor. The dictionary keys are symbols for the fitting parameters (e.g. "p").
+"""
+struct RandomizedBenchmarkingFitProperties
+    model_order::Union{Int, Nothing}
+    initial_parameters::Union{Dict, Nothing}
+
+    function RandomizedBenchmarkingFitProperties(model_order, initial_parameters)
```
alireza-najafi, line 18: All functions declarations should clearly specify arguments types. What is the model_order type? Is it Int?

src/benchmarking/randomized_benchmarking.jl
```diff
@@ -0,0 +1,342 @@
...
+                throw(ErrorException("there must be 5 initial parameters for the
+                    first-order fit"))
+            end
+        elseif !(model_order === nothing)
+            throw(ErrorException("the model order must be 0, 1, or nothing"))
+        end
+        new(model_order, initial_parameters)
+    end
+
+    function RandomizedBenchmarkingFitProperties(model_order)
```
alireza-najafi, line 35: Same comment as above.

src/benchmarking/randomized_benchmarking.jl
```diff
@@ -0,0 +1,342 @@
...
+    run_randomized_benchmarking(simulate_shots, properties::RandomizedBenchmarkingProperties, transpile! = (f(x)=x))
+
+Conducts randomized benchmarking and returns [`RandomizedBenchmarkingResults`](@ref).
+
+# Arguments
+- `simulate_shots`: a function which takes in an array of `QuantumCircuit` and returns an array of bit strings. The function should not transpile the circuits since it could transpile all the gates into a single identity operation. Use the transpile! argument instead.
+- `properties`: the [`RandomizedBenchmarkingProperties`](@ref) influencing benchmarking such as the list of target qubits and the list of sequence lengths.
+- `transpile!`: a function which transpiles a single circuit. It transpiles every Clifford operation individually.
+"""
+function run_randomized_benchmarking(simulate_shots,
```
alireza-najafi, line 139: What is the data type of similate_shots? Please specify in the function declaration. Don't simply rely on documentation.

src/benchmarking/randomized_benchmarking.jl
```diff
@@ -0,0 +1,342 @@
...
+    elseif num_lengths < 5 && order == 1
+        @warn "At least 5 sequence lengths are needed to generate a first-order fit! "*
+            "A zeroth-order fit will be used instead"
+        return RandomizedBenchmarkingFitProperties(0)
+    end
+    initial_parameters = properties.fit_properties.initial_parameters
+    return RandomizedBenchmarkingFitProperties(order, initial_parameters)
+end
+
+function get_sequence_fidelities(simulate_shots, transpile!,
```
alireza-najafi, line 167: same as above-

src/benchmarking/randomized_benchmarking.jl
```diff
@@ -0,0 +1,342 @@
...
+                properties.target_qubits)
+            circuit_id += 1
+        end
+        sequence_fidelity = survival_probability_sum/num_circuits_for_length
+        push!(sequence_fidelity_list, sequence_fidelity)
+    end
+    return sequence_fidelity_list
+end
+
+function get_survival_probability(shots, target_qubits)
```
alireza-najafi, line 194: same as above

src/benchmarking/randomized_benchmarking.jl
```diff
@@ -0,0 +1,342 @@
...
+        trimmed_measurement = measurement[target_qubits]
+        if trimmed_measurement == "0"^num_target_qubits
+            num_successes += 1
+        end
+    end
+    survival_probability = num_successes/length(shots)
+    return survival_probability
+end
+
+function get_random_clifford_circuits(sequence_length_id, transpile!,
```
alireza-najafi, line 207: same as above

src/benchmarking/randomized_benchmarking.jl
```diff
@@ -0,0 +1,342 @@
...
+        inverse_clifford = inv(clifford_product)
+        inverse_circuit = get_transpiled_circuit(inverse_clifford, properties,
+            qubit_map, transpile!)
+        append!(circuit, inverse_circuit)
+        push!(circuit_list, circuit)
+    end  
+    return circuit_list
+end
+
+function get_transpiled_circuit(clifford, properties, qubit_map, transpile!)
```
alireza-najafi, line 236: same as above. Signatures like this are very scary for another developer and hard to understand.

src/benchmarking/randomized_benchmarking.jl
```diff
@@ -0,0 +1,342 @@
...
+            "B1"=>fit.param[4], "C1"=>fit.param[5])
+        results = RandomizedBenchmarkingFitResults(model_order=order,
+            parameters=optimal_parameters,
+            residuals=fit.resid, jacobian=fit.jacobian,
+            converged=fit.converged)
+    end
+    return results
+end
+
+function get_zeroth_model(m, p)
```
alireza-najafi, line 282: What is the data type of m and p? There is absolutely no documentation here. Moreover, I noticed unit tests on my Mac fail on this function due to Segmentation Fault.

src/benchmarking/randomized_benchmarking.jl
```diff
@@ -0,0 +1,342 @@
...
+            optimal_parameters["A1"],
+            optimal_parameters["B1"],
+            optimal_parameters["C1"]]
+        plot_fidelity_model!(p, get_first_model, parameters_vector,
+            results.sequence_length_list)
+    end
+    return p
+end
+
+function plot_fidelity_model!(p, get_model, parameters_vector, sequence_length_list)
```
alireza-najafi, line 337: Same as above.

alireza-najafi (changes requested):
Nick, 
  I'm sending this back to you for revisions. Two important points: 

1. Please make sure you explicitly declare function argument types!! This would be very helpful for another developer to be able to understand your code. 


2. I see some Segmentation Fault issues when I run the unit tests on my Mac.

src/benchmarking/randomized_benchmarking.jl
npoirie1: I've been attempting to follow this recommendation from the Julia style guide: https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-writing-overly-specific-types. In cases where only a specific type is possible, I will add the type declaration.

src/benchmarking/randomized_benchmarking.jl
npoirie1: This function is not meant to be directly called by a user, which is why I haven't added documentation for it. The documentation for RandomizedBenchmarkingFitResults gives a description of each model including this one. I will add type declarations for m and p.

npoirie1:
This pull request also fixes #38.

src/benchmarking/randomized_benchmarking.jl
```diff
@@ -0,0 +1,351 @@
...
+randomized benchmarking calculation.
+
+The models for the fit are described in [`RandomizedBenchmarkingFitResults`](@ref). The
+fit can have `model_order` 0, 1, or `nothing`. If `nothing` is specified, no fit will be
+generated. Optionally, a dictionary of initial fitting parameters can be passed to the
+constructor. The dictionary keys are symbols for the fitting parameters (e.g. "p").
+"""
+struct RandomizedBenchmarkingFitProperties
+    model_order::Union{Integer, Nothing}
+    initial_parameters::Union{Dict{String, <:Real}, Nothing}
```
plerouxanyon, line 16: Why are the parameters in a `Dict` instead of being attributes of the class?

plerouxanyon (changes requested):
This is a very large change. I think the best would be for us to work through it iteratively until we covered everything.

src/benchmarking/randomized_benchmarking.jl
```diff
@@ -0,0 +1,351 @@
...
+- `num_bits_on_device::Int = 0`: the total number of classical bits on the device.
+- `target_qubits::Array{Int}`: a list of the qubits on which benchmarking is performed. By default, all qubits on the device are benchmarked.
+- `fit_properties`: the [`RandomizedBenchmarkingFitProperties`](@ref) which define the model for fitting the averaged sequence fidelity.
+- `sequence_length_list::Array{Int}`: a list indicating the number of Clifford operations in each circuit. For instance, specifying [1, 2] will lead to the generation of multiple circuits with 1 Clifford operation and multiple circuits with 2 Clifford operations.
+- `num_circuits_per_length::Array{Int}`: a list defining how many circuits to generate for each sequence length. By default, 100 circuits are generated for each sequence length.
+"""
+@with_kw struct RandomizedBenchmarkingProperties
+    num_qubits_on_device::Int; @assert num_qubits_on_device > 0
+    num_bits_on_device::Int = 0; @assert num_bits_on_device >= 0
+    target_qubits::Array{Int} = 1:num_qubits_on_device
```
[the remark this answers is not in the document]
npoirie1: The randomized benchmarking tool generates sequences of elements belonging to the Clifford group on n qubits, where n is the number of `target_qubits`.  A single element may contain multiple Clifford gates (including CNOTs). The benchmarking tool uses the same protocol as described by Magesan, Gambetta, and Emerson [http://dx.doi.org/10.1103/PhysRevA.85.042311].

Project.toml
```diff
@@ -8,6 +8,8 @@ Coverage = "a2441757-f6aa-5fb2-8edb-039e3f45d037"
 Interpolations = "a98d9a8b-a2ab-59e6-89dd-64a1c18fca59"
 JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+LsqFit = "2fda8390-95c7-5789-9bda-21331edee243"
+Nemo = "2edaba10-b0f1-5616-af89-8c11ac63239a"
```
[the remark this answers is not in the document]
npoirie1: All the `Nemo.jl` functions that are called should be easily replaceable by functions from `AbstractAlgebra.jl`, which contains no C dependencies. `Nemo.jl` uses C dependencies to provide optimized implementations of some of the functions in `AbstractAlgebra.jl`.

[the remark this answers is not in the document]
plerouxanyon: I've learned that we are not including the Manifest.toml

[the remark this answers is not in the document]
plerouxanyon: Let's try to stick to `AbstractAlgebra.jl` and look it it needs optimizing.

src/benchmarking/randomized_benchmarking.jl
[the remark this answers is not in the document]
plerouxanyon: Okay. I understand what it does.

It is also common to run parallel single-qubit benchmarking, which runs single-qubit benchmarking simultaneously on multiple qubits. This gives a more realistic number as cross-qubit effects are also considered. How do you recommend we extend what you have to allow for parallel benchmarking?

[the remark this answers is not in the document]
npoirie1: Parallel single-qubit benchmarking could be implemented by adding a `run_parallel_randomized_benchmarking` function. The function would take the same arguments as `run_randomized_benchmarking`. We could create a function called `get_parallel_random_clifford_circuits` which would be similar to the existing `get_random_clifford_circuits` function.  The new function would create single-qubit Clifford circuits and combine them such that the gates become parallel. The tools for generating the single-qubit Clifford circuits already exist. We would need to add a function that appends a circuit and makes the gates parallel if possible.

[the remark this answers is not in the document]
plerouxanyon: That seems like a good path forward and something we can add as an extension rather than implementing in this pull request.

PR #44, opened 2023-01-05
Feat/add measurement probabilities

### LINKED ISSUE #43: Add a function which, given a ket, returns measurement probabilities

### LINKED ISSUE #59: simulate_shots does not correctly handle complex values
The amplitudes computed in the simulate_shots function are incorrect when kets have nonzero imaginary terms.

For instance, the samples in the following example should be "0" instead of "1":

```
julia> circuit = QuantumCircuit(qubit_count=1, bit_count=0);

julia> push_gate!(circuit, sigma_x(1));

julia> push_gate!(circuit, phase(1));

julia> push_gate!(circuit, sigma_x(1));

julia> push_gate!(circuit, phase(1))
Quantum Circuit Object:
   id: 52530f1c-b6d9-11ed-0fca-b9527064440e 
   qubit_count: 1 
   bit_count: 0 
q[1]:──X────S────X────S──
                         



julia> shots = simulate_shots(circuit, 5)
5-element Vector{String}:
 "1"
 "1"
 "1"
 "1"
 "1"
```

npoirie1:
Adds functions for computing the measurement probabilities of a ket or a circuit. Probabilities for a subset of qubits can also be determined.

Fixes #43 and fixes #59.

### DIFF
--- a/src/core/qobj.jl
+++ b/src/core/qobj.jl
@@ -709,30 +709,35 @@ function normalize!(x::Ket)
     return x
 end
 
-# function get_measurement_probabilities(x::Ket)
-#     return real.(x .* x)
-# end
+function get_measurement_probabilities(x::Ket)
+    return real.(x .* x)
+end
 
-# function get_measurement_probabilities(x::Ket, target_bodies::Vector{<:Integer},
-#     hilbert_space_size_per_body::Integer=2)
+function get_measurement_probabilities(x::Ket, target_bodies::Vector{<:Integer},
+    hilbert_space_size_per_body::Integer=2)
 
-#     amplitudes = real.(x .* x)
-#     num_amplitudes = length(amplitudes)
-#     num_target_amplitudes = hilbert_space_size_per_body^length(target_bodies)
-#     if num_target_amplitudes > num_amplitudes
-#         throw(ErrorException("too many target bodies were provided"))
-#     elseif num_target_amplitudes == num_amplitudes
-#         return get_measurement_probabilities(x)
-#     else
-#         target_amplitudes = Vector{Float64}(undef, num_target_amplitudes)
-#         num_summed_amplitudes = num_amplitudes-num_target_amplitudes
-#         for i_target_amplitude in 0:num_target_amplitudes-1
-#             for i_summed_amplitude in 0:num_summed_amplitudes-1
-
-#             end
-#         end
-#     end
-# end
+    amplitudes = real.(x .* x)
+    num_amplitudes = length(amplitudes)
+    num_target_amplitudes = hilbert_space_size_per_body^length(target_bodies)
+    if num_target_amplitudes == num_amplitudes
+        return get_measurement_probabilities(x)
+    else
+        num_bodies = get_num_bodies(x, hilbert_space_size_per_body)
+        remaining_bodies = [x for x ∈ 1:num_bodies if x ∉ target_bodies]
+        target_amplitudes = Vector{Float64}(undef, num_target_amplitudes)
+        num_summed_amplitudes = hilbert_space_size_per_body^length(remaining_bodies)
+        for i_target_amplitude in 0:num_target_amplitudes-1
+            sum = 0
+            for i_summed_amplitude in 0:num_summed_amplitudes-1
+                ket_index = get_ket_index(target_bodies, remaining_bodies,
+                    hilbert_space_size_per_body, i_target_amplitude, i_summed_amplitude)
+                sum += amplitudes[ket_index]
+            end
+            target_amplitudes[i_target_amplitude+1] = sum
+        end
+        return target_amplitudes
+    end
+end
 
 function get_ket_index(target_bodies::Vector{<:Integer},
     remaining_bodies::Vector{<:Integer},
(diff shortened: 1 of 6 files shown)


npoirie1:
I found a bug with the get_measurement_probabilities function. When an entry in a ket has real and imaginary parts that are negative, the computed probability is negative. I will fix this issue immediately.

npoirie1:
The bug in the get_measurement_probabilities functions has been fixed. A similar bug in the simulate_shots function has also been fixed.

test/qobj.jl
```diff
@@ -143,4 +143,57 @@ end
...
+end
+
+@testset "change_to_decimal_integer" begin
+    symbols = [0, 1, 2, 1]
+    base = 3
+    decimal = Snowflake.change_to_decimal_integer(symbols, base)
+    @test decimal == 16
+end
+
+@testset "get_ket_index" begin
```
plerouxanyon, line 170: See comment of `change_number_base` test

src/core/qobj.jl
```diff
@@ -661,6 +661,138 @@ function normalize!(x::Ket)
...
+julia> target_qubit = [2];
+
+julia> get_measurement_probabilities(ψ, target_qubit)
+2-element Vector{Float64}:
+ 0.9999999999999998
+ 0.0
+
+```
+"""
+function get_measurement_probabilities(x::Ket)
```
plerouxanyon, line 710: Specifying the return type would be useful for a user. It is not ideal if a user should run or look at the code to determine the function's return type.

This applies to all `get_measurement_probabilities` functions.

src/core/qobj.jl
```diff
@@ -661,6 +661,138 @@ function normalize!(x::Ket)
...
+    remaining_bodies::Vector{<:Integer},
+    hspace_size_per_body::Integer, target_index::Integer,
+    remainder_index::Integer)
+
+    target_symbols = change_number_base(target_index, hspace_size_per_body,
+        length(target_bodies))
+    remainder_symbols = change_number_base(remainder_index, hspace_size_per_body,
+        length(remaining_bodies))
+    num_symbols = length(target_bodies)+length(remaining_bodies)
+    combined_symbols = Vector{Int}(undef, num_symbols)
```
plerouxanyon, line 750: Using concrete type, `Int`, but dispatching on any `Integer` (`Int8`, `Int16`, ...)

plerouxanyon (changes requested):

plerouxanyon:
Regarding return type annotation in Julia:

@npoirie1  has sent me a link: https://discourse.julialang.org/t/do-you-put-return-type-in-function-definitions/32120/16

There doesn't appear to be a consensus. In this specific case:

* Ket is a generic type and not an abstract type
* The type result is obvious given the type input
* Annotating the return type with `AbstractVector{T}` would not limit the implementation

Considering all this, I think that, in this case, it makes sense to annotate the return type as part of the public interface. However, the return type should still be as abstract as reasonable (`AbtractVector` and not `Vector`, etc.). This won't be limiting, and changing the return type significantly would be a breaking change anyway. Having the return type available is really useful for allowing code to be reasoned about locally without diving into another function's implementation detail.

I think the best would be to handle this on a case-by-case basis.

test/qobj.jl
```diff
@@ -143,4 +143,57 @@ end
 @testset "is_hermitian" begin
     @test is_hermitian(sigma_y())
     @test !is_hermitian(sigma_p())
-end
\ No newline at end of file
+end
+
+@testset "change_number_base" begin
```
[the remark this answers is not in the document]
npoirie1: The tests of "private" functions have been removed.

test/qobj.jl
npoirie1: The tests of "private" functions have been removed.

test/qobj.jl
```diff
@@ -143,4 +143,57 @@ end
...
+    decimal = 88
+    @test_throws ErrorException Snowflake.change_number_base(decimal, base, num_symbols)
+
+    decimal = 16
+    symbols = Snowflake.change_number_base(decimal, base, num_symbols)
+    expected_symbols = [0, 1, 2, 1]
+    @test symbols == expected_symbols
+end
+
+@testset "change_to_decimal_integer" begin
```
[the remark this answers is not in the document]
npoirie1: The tests of "private" functions have been removed.

src/core/qobj.jl
npoirie1: Now using the parametric type of the arguments.

npoirie1:
@plerouxanyon, all your comments have been addressed. The get_measurement_probabilities function now accepts a different Hilbert space size for each body.

src/core/qobj.jl
```diff
@@ -682,6 +684,159 @@ function normalize!(x::Ket)
...
+    hspace_size_per_body_list = fill(hspace_size_per_body, num_bodies)
+    return get_measurement_probabilities(x, target_bodies, hspace_size_per_body_list)
+end
+
+function get_measurement_probabilities(x::Ket{Complex{T}},
+    target_bodies::Vector{U},
+    hspace_size_per_body::Vector{U})::AbstractVector{T} where {T<:Real, U<:Integer}
+
+    throw_if_targets_are_invalid(x, target_bodies, hspace_size_per_body)
+    amplitudes = real.(adjoint.(x) .* x)
```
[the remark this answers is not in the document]
npoirie1: `get_measurement_probabilities` is now reused

npoirie1:
@plerouxanyon, the new comment and the merge conflicts have been addressed.

PR #57, opened 2023-02-24
Fix/add strict doctests

### LINKED ISSUE #56: CI/CD check for documentation passes when doctests fail
The CI/CD check "Documentation / build (pull_request)" passes even when the doctests fail. As a solution, the strict keyword in makedocs could be set to true or :doctest. Setting this keyword should cause the build procedure to fail when doctests fail. For more details, see: https://docs.juliahub.com/Documenter/sHGjf/0.27.18/lib/public/#Documenter.makedocs.

npoirie1:
This pull request should cause the CI/CD to fail for the documentation if there is an error in the doctests. This was not the case before. All errors and warnings were also removed from the documentation. I will run a test to ensure that the CI/CD does indeed fail if there is a doctest error.

Fixes #56

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -15,7 +15,8 @@ makedocs(
         #"Simulating Quantum Systems" => "simulating_quantum_systems.md",            
         "Library" => "library.md",
     ],
-    doctestfilters = [uuid_regex]
+    doctestfilters = [uuid_regex],
+    strict = true
 )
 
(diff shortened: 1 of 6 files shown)


npoirie1:
As planned, the documentation CI/CD fails if the doctests fail. I will remove the error.

plerouxanyon (approved):
Looks good. I'll merge when the CI passes.

PR #60, opened 2023-03-02
fix: gate constructor only accept integer targets, and the target cou…

fSosaRey:
fix: enforce target count and type


Project.toml
```diff
@@ -5,6 +5,7 @@ version = "0.1.0"
 
 [deps]
 Coverage = "a2441757-f6aa-5fb2-8edb-039e3f45d037"
+Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
```
plerouxanyon, line 8: Why is this dependency needed?

Project.toml
```diff
@@ -15,6 +16,7 @@ Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 ProtoBuf = "3349acd9-ac6a-5e09-bcdb-63829b23a429"
 Revise = "295af30f-e4ad-537b-8983-00126c2a3abe"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
```
plerouxanyon, line 19: A dependency was added, but the `Manifest.toml` was not updated.

plerouxanyon (changes requested):

plerouxanyon (approved):
Looks good

PR #71, opened 2023-03-14
Feat/abstract operator prototype

fSosaRey:
Phase and T gate implemented with DiagonalOperator. Still need to fix codecov.


src/core/quantum_gate.jl
```diff
@@ -97,6 +97,20 @@ Underlying data Matrix{ComplexF32}:
 """
 abstract type Gate end
 
+abstract type AbstractGate end
+
+nonexistent_gate(target::Integer) = NonExistentGate(target) # to test MethodError on non-implemented AbstractGates
```
plerouxanyon, line 102: This should be part of the tests and not part of the library

src/core/quantum_gate.jl
```diff
@@ -147,6 +161,52 @@ function apply_gate!(state::Ket, gate::Gate)
...
+    throw(NotImplementedError(:get_connected_qubits, gate))
+
+
+function apply_operator!(state::Ket,operator::DiagonalOperator,connected_qubit::Int,qubit_count::Int)
+
+    dim=2^qubit_count
+    target_qubit_index=qubit_count-connected_qubit # indexing of targets in qulacs starts at 0
+    
+    diagonal_in_matrix=operator.data
+
```
plerouxanyon, line 195: This implementation assumes a single qubit operator. We need to generalize it or mark it as a specialization.

Comparatively, what performance do we get when implementing the diagonal operator using `@simd` in the loop?

src/core/quantum_gate.jl
```diff
@@ -666,6 +746,26 @@ get_operator(gate::Pi8) = pi_8(gate.type)
...
 
+"""
+    pi_8_diag(target)
+
+Return a π/8 `Gate` (also known as a ``T`` `Gate`), which applies the [`pi_8_diag()`](@ref) `DiagonalOperator` to the `target` qubit.
+"""
+pi_8_diag(target::Integer) = Pi8_Diag(["T"], "t", target)
+
+struct Pi8_Diag <: AbstractGate
+    display_symbol::Vector{String}
```
plerouxanyon, line 757: Display symbol should not be part of the gate. It should be part of what is displaying the gate.

src/core/quantum_gate.jl
```diff
@@ -666,6 +746,26 @@ get_operator(gate::Pi8) = pi_8(gate.type)
...
+    instruction_symbol::String
```
plerouxanyon, line 758: The instruction symbol should not be part of the gate. It should be part of the QPU the gate is being sent.

src/core/qobj.jl
```diff
@@ -437,6 +467,32 @@ function Base.show(io::IO, x::Operator)
     end
 end
 
+function Base.show(io::IO, x::DiagonalOperator)
+    println(io, "$(size(x.data))-element Snowflake.DiagonalOperator:")
+    println(io, "Underlying data $(typeof(x.data)):")
```
plerouxanyon, line 472: The underlying data should be an implementation detail and not part of the public interface

src/core/qobj.jl
```diff
@@ -119,12 +119,41 @@ struct Operator{T<:Complex}
...
+julia> z = Snowflake.DiagonalOperator([1.0,-1.0])
+(2,)-element Snowflake.DiagonalOperator:
+Underlying data StaticArraysCore.SVector{2, ComplexF64}:
+1.0 + 0.0im    0.0 + 0.0im
+0.0 + 0.0im    -1.0 + 0.0im
+
+```
+"""
+struct DiagonalOperator{T<:Complex}<:AbstractOperator
+    data::SVector{2,T}
```
plerouxanyon, line 147: This needs to be generalized or marked as a specialization.

plerouxanyon (changes requested):
We must also remove the `diag` suffix when moving to the new gate abstraction.

Also, please rebase this branch against the main when merging the benchmarking branch.

plerouxanyon:
Everything looks good, except plotting fails if you did not run all the benchmarks. How would one benchmark only a single gate and compare results with other benchmarks?

src/core/qobj.jl
```diff
@@ -119,12 +119,51 @@ struct Operator{T<:Complex}
...
+Underlying data type: ComplexF64:
+1.0 + 1.0im    0.0 + 0.0im    0.0 + 0.0im    0.0 + 0.0im
+0.0 + 0.0im    1.0 + 0.0im    0.0 + 0.0im    0.0 + 0.0im
+0.0 + 0.0im    0.0 + 0.0im    1.0 + 0.0im    0.0 + 0.0im
+0.0 + 0.0im    0.0 + 0.0im    0.0 + 0.0im    0.0 - 1.0im
+
+```
+"""
+struct DiagonalOperator{N,T<:Complex}<:AbstractOperator
+    data::SVector{N,T}
```
plerouxanyon, line 155: What is the performance difference when we use a vector instead of a static vector?

src/core/quantum_gate.jl
```diff
@@ -147,6 +155,145 @@ function apply_gate!(state::Ket, gate::Gate)
...
+            "not enough qubits in the Ket for the targets in gate"))
+    end
+
+    connected_qubits=get_connected_qubits(gate)
+
+    type_in_ket=eltype(state.data)
+
+    operator=get_operator(gate,type_in_ket)
+
+    apply_operator!(state,operator,connected_qubits,Int(qubit_count),Val(length(connected_qubits)))
```
plerouxanyon, line 176: The operator already has the size fixed at compile time. Is it necessary to pass the number of connected qubits as a parameter?

src/core/quantum_gate.jl
```diff
@@ -147,6 +155,145 @@ function apply_gate!(state::Ket, gate::Gate)
...
+function apply_operator!(
+    state::Ket,
+    operator::DiagonalOperator,
+    connected_qubits::Vector{<:Integer},
+    qubit_count::Int,
+    ::Val{N_targets}) where {N_targets} #specialization for N_targets>1 diagonal gates
+
+    dim=2^qubit_count
+    
+    target_qubit_index_list=Vector{Int64}([qubit_count-t for t in reverse(connected_qubits)]) # the bitwise implementation assumes target numbering starting at 0
```
plerouxanyon, line 260: This line becomes a bit long with the inline comment. Perhaps we should move it above the line.

src/core/quantum_gate.jl
```diff
@@ -147,6 +155,145 @@ function apply_gate!(state::Ket, gate::Gate)
...
+    end
+
+end
+
+function apply_operator!(
+    state::Ket,
+    operator::DiagonalOperator,
+    connected_qubits::Vector{<:Integer},
+    qubit_count::Int,
+    ::Val{N_targets}) where {N_targets} #specialization for N_targets>1 diagonal gates
```
plerouxanyon, line 256: Do we gain any performance when templating the number of targets?

src/core/quantum_gate.jl
```diff
@@ -666,6 +833,24 @@ get_operator(gate::Pi8) = pi_8(gate.type)
...
 
+"""
+    pi_8_diag(target)
+
+Return a π/8 `Gate` (also known as a ``T`` `Gate`), which applies the [`pi_8_diag()`](@ref) `DiagonalOperator` to the `target` qubit.
+"""
+pi_8_diag(target::Integer) = Pi8_Diag([target])
+
+struct Pi8_Diag <: AbstractGate
+    target_list::Vector{<:Integer}
```
plerouxanyon, line 844: 1. There is only one target so lets not have the `Vector`.
2. This can simply be an `Int` unless there is a benefit to having it templated?

src/core/quantum_gate.jl
```diff
@@ -788,6 +973,20 @@ get_operator(gate::RotationZ) = rotation_z(gate.parameters[1],gate.type)
 
 get_inverse(gate::RotationZ) = rotation_z(gate.target[1], -gate.parameters[1],gate.type)  
 
+phase_shift_diag(target::Integer, phi::Real) = PhaseShift_Diag([target], phi)
+
+struct PhaseShift_Diag <: AbstractGate
+    target_list::Vector{<:Integer}
```
plerouxanyon, line 979: See above comment

test/diagonal.jl
```diff
@@ -0,0 +1,205 @@
...
+    T_gate_error=Snowflake.pi_8_diag(target_error)
+
+    @test_throws DomainError apply_gate!(ψ, T_gate_error)
+
+    @test_throws NotImplementedError get_inverse(T_gate_error) #TODO, implement inverse of DiagonalGate
+
+    ######################################################
+
+
+    nonexistent_gate(target::Integer) = NonExistentGate(target) # to test MethodError on non-implemented AbstractGates
```
plerouxanyon, line 107: This is a test for AbstractGate and not for DiagonalOperate/Gate

test/diagonal.jl
```diff
@@ -0,0 +1,205 @@
+using Snowflake
+using Test
+using StaticArrays
+
+@testset "diagonal_gate" begin
```
plerouxanyon, line 5: This test group is getting a bit large. I think we should split it into pi8_gate and phase_gate

src/core/quantum_gate.jl
```diff
@@ -666,6 +833,24 @@ get_operator(gate::Pi8) = pi_8(gate.type)
...
+"""
+pi_8_diag(target::Integer) = Pi8_Diag([target])
+
+struct Pi8_Diag <: AbstractGate
+    target_list::Vector{<:Integer}
+end
+
+get_operator(gate::Pi8_Diag,T::Type{<:Complex}=ComplexF64) = pi_8_diag(T)
+
+get_inverse(gate::Pi8_Diag) =  throw(NotImplementedError(:get_inverse, gate)) #TODO
```
plerouxanyon, line 849: Just leaving a comment here so we remember that we need to implement this.

plerouxanyon (changes requested):

PR #73, opened 2023-03-20
Feat/diagonal operator and gates

fSosaRey:
feat: gates with diagonal operators fully integrated with existing features and tests


plerouxanyon (changes requested):

test/diagonalGate.jl
```diff
@@ -0,0 +1,220 @@
+using Snowflake
+using Test
+using StaticArrays
+
+@testset "Diagonal Gate: phase_shift" begin
+
```
plerouxanyon, line 6: I would like to refactor the tests a bit.

Testsets allow grouping and naming tests.

Can we try the following structure?

* "Phase shift gate"
  * "Apply gate is equivalent to apply phase shift operator"
  * "Constructor"
  * "Apply gate applied on the last qubit"
  * "Apply gate applied on a non-last qubit"
  * "Adjoint"
  * "Get inverse"
  * "Legacy" (Useful to know which functionality won't be there in the final version.)
  
 Adapt as it is necessary, but then it is very clear what we are testing, and then we don't need comments for any test groups. Naturally, this also applies to the rest of the tests for "new" gates.

test/testgroups
```diff
@@ -5,4 +5,6 @@ dynamic_system
 circuit
 gate
 transpile
-submit_circuit
\ No newline at end of file
+submit_circuit
```
plerouxanyon, line 8: Wasn't `submit_circuit` removed?

plerouxanyon (changes requested):

test/testgroups
fSosaRey: only the call to `push_gate!(c, [sigma_x(1)])` was commented, as the push cannot yet handle AbstractGates

PR #110, opened 2023-03-27
Feat/abstract dense gates

fSosaRey:


src/core/quantum_gate.jl
```diff
@@ -921,21 +969,21 @@ Return a gate that applies a rotation `theta` to the `target` qubit about the co
...
-    [theta, phi], T)
+rotation(target::Integer, theta::Real, phi::Real) = Rotation(target, [theta, phi])
 
-struct Rotation <: Gate
-    display_symbol::Vector{String}
-    instruction_symbol::String
-    target::SVector{1,Int}
+struct Rotation <: AbstractGate
+    target::Int
     parameters::SVector{2,Real}
```
plerouxanyon, line 976: Rather name the parameters explicitly. It will be much clearer which one is theta and which one is phi.

src/core/quantum_gate.jl
```diff
@@ -1073,16 +1105,11 @@ function ensure_target_qubits_are_different(target::Array)
     end
 end
 
-struct ControlZ <: Gate
-    display_symbol::Vector{String}
-    instruction_symbol::String
+struct ControlZ <: AbstractGate
     target::SVector{2,Int}
```
plerouxanyon, line 1109: Have on `target` and one `control`  for the respective qubits.

src/core/quantum_gate.jl
```diff
@@ -1115,22 +1137,19 @@ Return the imaginary swap `Gate` which applies the imaginary swap `Operator` to
...
     ensure_target_qubits_are_different(target)
-    return ISwap(["x", "x"], "iswap", target, T)
+    return ISwap(target)
 end
 
-struct ISwap <: Gate
-    display_symbol::Vector{String}
-    instruction_symbol::String
+struct ISwap <: AbstractGate
     target::SVector{2,Int}
```
plerouxanyon, line 1147: This is plural. It should be `targets`

test/diagonalGate.jl
```diff
@@ -154,3 +154,93 @@ end
...
 
+@testset "Diagonal Gate: N targets" begin
+
+
+    ######################################################
+
+    diagonalGate_N_targets(target_list::Vector{<:Integer}) = 
+        Diagonal_N_TargetGate(target_list)
+
+    struct Diagonal_N_TargetGate <: Snowflake.AbstractGate
```
plerouxanyon, line 165: If there are no multi-qubit diagonal gates, rather remove the implementation than define a gate here.

test/gate.jl
```diff
@@ -251,13 +235,12 @@ end
...
-        instruction_symbol::String
-        target::Vector{Int}
-        parameters::Vector{Int}
+    struct UnknownGateWithoutOperator <: AbstractGate
+        target::Int
+        parameter::Int
     end
 
-    unknown_gate = UnknownGateWithoutOperator("na", [1], [])
+    unknown_gate = UnknownGateWithoutOperator(1, 1)
```
plerouxanyon, line 243: What is being tested here?

src/core/qobj.jl
```diff
@@ -315,80 +284,66 @@ false
 
 ```
 """
-is_hermitian(A::Operator) = LinearAlgebra.ishermitian(A.data)
-
-is_hermitian(A::AbstractOperator)    = LinearAlgebra.ishermitian(Operator(A).data)
+is_hermitian(A::AbstractOperator) = LinearAlgebra.ishermitian(DenseOperator(A).data)
```
plerouxanyon, line 287: Why are we redefining this instead of using `ishermitian`?

plerouxanyon (changes requested):

test/gate.jl
fSosaRey: Was part of main, but nothing is covered by it. Removing.

plerouxanyon:
I noticed that the `PhaseShift` and the `RotationZ` implement the same gate. I think we should remove the phase shift gate.

plerouxanyon:
Also, the `Pi8` gate does not correspond to the canonical name of the gate it is implementing. I think we should rename it to the `T` gate.

plerouxanyon:
This is superseded by #112

Issue #85, opened 2023-03-21
Implement Anyon Quantum Computer QPU interface


plerouxanyon:
Use client #84 to implement the QPU interface #83.

Issue #84, opened 2023-03-21
Add Anyon Quantum Computer Client


plerouxanyon:
A low-level interface to Anyon's quantum computer

The client will be able to submit jobs, get the job status, and get job results

Issue #83, opened 2023-03-21
Add QPU interface


plerouxanyon:
The interface will abstract away details of the underlying quantum computer so that experiments do not have to depend on a concrete device.

plerouxanyon:
This depends on #82

PR #115, opened 2023-04-04
Feat/qpu interface

fSosaRey:
minimal qpu-interface


src/remote/run_jobs.jl
```diff
@@ -0,0 +1,198 @@
...
+        printout_response(:get_result,formatted_response,typeof(response))
+    end
+
+    return body["histogram"]
+end
+
+
+abstract type Service end
+
+struct QPUService <: Service
```
plerouxanyon, line 135: Please rename to `AnyonQPU`

It should also only depend on client. It is going to replace the old `QPU`.

src/remote/run_jobs.jl
```diff
@@ -0,0 +1,198 @@
...
+        end
+
+        new(client,qpu)
+    end
+end
+
+get_client(qpu_service::QPUService)=qpu_service.client
+get_qpu(qpu_service::QPUService)=qpu_service.qpu
+
+function Base.Dict(response::HTTP.Messages.Response)
```
plerouxanyon, line 155: This is type piracy and could break external code if they import this package

See https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-type-piracy.

src/remote/run_jobs.jl
```diff
@@ -0,0 +1,198 @@
...
+    )
+
+    # convert response body from binary to ASCII
+    read_buffer=IOBuffer(reinterpret(UInt8, response.body))
+    formatted_response["body"]=String(readuntil(read_buffer, 0x00))
+
+    return formatted_response
+end
+
+function run(qpu_service::QPUService, circuit::QuantumCircuit,num_repetitions::Integer;verbose=false)
```
plerouxanyon, line 170: Could you remove the verbose flag? Logging should be done using middle layers and composition, not by passing flags into functions.

src/remote/run_jobs.jl
```diff
@@ -0,0 +1,198 @@
...
+
+        if verbose
+            println("status: $(status)")
+        end
+        
+        if !(status in ["queued","running"])
+            break
+        end
+
+        sleep(0.2) #wait 200ms to minimize printout counts
```
plerouxanyon, line 192: Would you please make this a parameter of `AnyonQPU` instead?

src/remote/run_jobs.jl
```diff
@@ -0,0 +1,198 @@
...
+struct Client 
+    host::String
+    user::String
+    access_token::String
+end
+
+get_host(client::Client)= client.host
+
+# submits circuit to host and returns circuitID
+function submit_circuit(client::Client,circuit_json::String;verbose=false)
```
plerouxanyon, line 47: Submit circuit should take a circuit and number of repetitions. The fact the client serialized the circuit to JSON is an implementation detail.

src/remote/run_jobs.jl
```diff
@@ -0,0 +1,198 @@
+using Snowflake
+using JSON
+using HTTP
+
+commonpath="benchmarking/data"
+
+function serialize_circuit(circuit::QuantumCircuit,repetitions::Integer;indentation::Integer=4)
```
plerouxanyon, line 7: You should never indent JSON that is sent over the network. If you want to be able to debug this perhaps return `circuit_description` instead and call `JSON.json(...)` when making the request in `submit_circuit`.

plerouxanyon (changes requested):
Let's split this MR, one for each of the issues

* #84 
* #83 
* #85

plerouxanyon:
Please check the commits that are part of this pull request. It shows changes that have already landed in `main`.

src/anyon/run_jobs.jl
```diff
@@ -0,0 +1,490 @@
...
+            "Content-Type"=>"application/json"
+            ),
+    )
+end
+
+function get_request(
+    ::MockRequestor,
+    url::String,
+    access_token::String
+    )
```
plerouxanyon, line 76: Add the return type

src/anyon/run_jobs.jl
```diff
@@ -0,0 +1,490 @@
...
+        get_requestor(client),
+        path_url,
+        get_access_token(client)
+        )
+
+    formatted_response=format_response(response)
+
+    body=JSON.parse(formatted_response["body"])
+
+    return body["status"]
```
plerouxanyon, line 267: Validate that the status is one of the expected status values.

src/anyon/run_jobs.jl
```diff
@@ -0,0 +1,490 @@
...
+
+    while true
+        
+        println("status: $(status["type"])")
+        
+        if !(status["type"] in ["queued","running"])
+            break
+        end
+
+        sleep(get_printout_delay(qpu)/1000)
```
plerouxanyon, line 445: Let's not make the wait between statuses dependent on print duration. We could want the latest status to be printed only every second but query the status every 100ms.

src/anyon/run_jobs.jl
```diff
@@ -0,0 +1,490 @@
...
+        end
+
+        sleep(get_printout_delay(qpu)/1000) 
+        status=get_status(client,circuitID;)
+    end
+    
+    if status["type"] == "failed"       
+        return Dict("error_msg"=>status["message"])
+    else
+        return get_result(client,circuitID)
```
plerouxanyon, line 452: Don't assume that the status is now success. It might be `cancelled`.

plerouxanyon (changes requested):

src/anyon/run_jobs.jl
```diff
@@ -0,0 +1,490 @@
...
+
+abstract type Requestor end
+
+get_request(requestor::Requestor,::String,::String,::String) = 
+    throw(NotImplementedError(:get_request,requestor))
+post_request(requestor::Requestor,::String,::String,::String) = 
+    throw(NotImplementedError(:post_request,requestor))
+
+struct HTTPRequestor<:Requestor end
+struct MockRequestor<:Requestor end
```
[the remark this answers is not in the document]
fSosaRey: The doctests require the MockRequester in the library code

plerouxanyon (changes requested):

plerouxanyon (changes requested):

src/anyon/run_jobs.jl
```diff
@@ -0,0 +1,549 @@
...
+
+julia> run_job(qpu,QuantumCircuit(qubit_count=3,gates=[sigma_x(3),control_z(2,1)]) ,100)
+Status: succeeded
+
+Dict{String, Int64} with 1 entry:
+  "001" => 100
+
+```
+"""
+function run_job(qpu::AnyonQPU, circuit::QuantumCircuit,num_repetitions::Integer)::Dict{String,Int}
```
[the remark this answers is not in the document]
fSosaRey: no printout is produced in run_job()

test/qpu_interface.jl
```diff
@@ -0,0 +1,168 @@
...
+
+    circuitID="1234-abcd"
+
+    response=get_request(
+        MockRequestor(),
+        joinpath(host,Snowflake.path_circuits,circuitID),
+        access_token
+    )
+
+    for f in fieldnames(typeof(response))
```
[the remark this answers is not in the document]
plerouxanyon: Also, please only mark it as resolved after pushing the commit (refactoring response comparison)

src/anyon/run_jobs.jl
```diff
@@ -0,0 +1,549 @@
...
+    printout_delay=get_printout_delay(qpu)/1e6 #convert ms to ns
+ 
+    ref_time_query=Base.time_ns()
+    query_delay=100/1e6 #100 ms between queries to host
+
+    while true        
+        current_time=Base.time_ns()
+
+        if (current_time-ref_time_print)>printout_delay
+            println(status)
```
plerouxanyon, line 489: Here is still a printout. Please only mark threads as resolved once a commit has been pushed.

test/qpu_interface.jl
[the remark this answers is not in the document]
fSosaRey: Version is the HTTP protocol version used to contruct the response

plerouxanyon (changes requested):
image

When I generate the documentation, the setup code appears in the docs. I expected the setup code to be hidden. Would you please check this?

Other than the setup code in the docs everything else is good to go!

plerouxanyon (approved):

Issue #132, opened 2023-04-18
CNot inverse identity does not hold


plerouxanyon:
The following identity does not hold

![image](https://user-images.githubusercontent.com/99685336/232833318-5e1af36a-e995-4d94-bee4-d49d5ecee9b7.png)
From Wikipedia: https://en.wikipedia.org/wiki/Controlled_NOT_gate

PR #133, opened 2023-04-18
Feat/place qubits onto a line

fSosaRey:
## Blocked by

- [x] #132


src/core/transpile.jl
```diff
@@ -479,3 +484,126 @@ function transpile(::CastToPhaseShiftAndHalfRotationX, circuit::QuantumCircuit):
...
+            targets_indices=findall(x->x in targets,connected_qubits)
+    
+            # left untouched
+            min_target=connected_qubits[1] 
+
+            compressed_targets=[min_target+i for i in 0:(length(connected_qubits)-1)]
+
+            previous_min_target=min_target
+
+            # for controlled gates, target qubit is last argument
```
plerouxanyon, line 570: This is not necessarily true.

src/core/transpile.jl
```diff
@@ -479,3 +484,126 @@ function transpile(::CastToPhaseShiftAndHalfRotationX, circuit::QuantumCircuit):
...
+
+    for gate in gates
+
+        connected_qubits=sort(get_connected_qubits(gate))
+
+        if length(connected_qubits)>1
+
+            targets=get_targets(gate)
+    
+            # in controlled gates, length(target_indices)<length(connected_qubits)
```
plerouxanyon, line 560: This transpiler should not be concerned with controlled gates.

We can have another optimizing transpiler which removes swap gates that re-order control qubits.

src/core/transpile.jl
```diff
@@ -479,3 +484,126 @@ function transpile(::CastToPhaseShiftAndHalfRotationX, circuit::QuantumCircuit):
...
+
+            @assert get_targets(gates_block[1])==compressed_targets[targets_indices] (
+                "Failed to construct gate: $(typeof((gates_block[1])))")
+
+            compressed_targets=reverse(compressed_targets)
+
+            for (i,next_target) in enumerate(reverse(connected_qubits[2:end]))
+                nearest_target=compressed_targets[i]
+
+                while !isequal(next_target,nearest_target)
```
plerouxanyon, line 585: This function is getting complex. We could factor out a qubit remap function which takes a new qubit order and adds swaps to get there. For example

`multi_swap_qubits!(circuit::QuantumCircuit, qubit_mapping::AbstractVector{Int})`

for example with
`circuit = circuit::QuantumCircuit(qubit_count=3, gates=[...])`

when
`multi_swap_qubits(circuit, [3,1,2])`

will add swaps to `circuit` so that qubit 3 is now at position 1, qubit 1 is now at 2, and qubit 2 is not at position 3.

Does it make sense what I am trying to describe?

test/transpile.jl
```diff
@@ -178,7 +178,46 @@ end
...
+
+    test_specs=[
+        # (gates list       gates_in_output)
+        ([control_z(5,6)],  1)      # no effect
+        ([control_z(5,10)], 9)      # target at bottom
+        ([control_z(10,1)], 17)     # target on top
+
+        ([toffoli(4,5,6)],  1)      # no effect
+        ([toffoli(4,6,5)],  1)      # no effect
+        # ([toffoli(6,5,4)],  1)      # no effect # fails, related to issue #132
```
plerouxanyon, line 195: Uncomment

plerouxanyon (changes requested):

src/core/transpile.jl
fSosaRey: This transpiler stage need to distinguish between control qubits and target qubits because when constructing the substitute with `typeof(gate)(connected_qubits_output...)` target is the last argument as per the controlled gates ctor. Both control and targets can be remapped, but the constructor arguments require correct ordering of target vs controls

fSosaRey: unless we convert all controlled gates constructors to `@kwdef`

plerouxanyon (changes requested):

src/core/quantum_circuit.jl
```diff
@@ -453,7 +453,7 @@ end
 function get_split_circuit_layout(io::IO, circuit_layout::Array{String},
     padding_width::Integer)
 
-    (display_height, display_width) = displaysize(io)
+    (display_height, display_width) = displaysize(stdout)
```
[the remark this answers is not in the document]
fSosaRey: this only affects the doctests, in which case there's a difference between the outputs of `displaysize(io)` and `displaysize(stdout)`, the former always returning the default width of 80. The other instances of `::IO` are `stderr` and `stdin`, which won't be passed by `Base.show()`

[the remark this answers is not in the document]
plerouxanyon: `IO` is an interface and can be much more than just `stderr` and `stdin`. It can even be user-defined, and we can easily get an `IO` type with another `displaysize` which will render incorrectly with base types.

```julia
    buffer = IOBuffer(UInt8[], read=true, write=true)
    io = IOContext(:displaysize = (30, 120))
```

We really should use the `displaysize` of the `IO` type. If the doctests are inconsistent with their width then we need to find a way to fix the inconsistency. If the doctests are consistent and always use the same width then the output of the doctests should be updated to match the consistent behaviour of doctests.

plerouxanyon (approved):

PR #140, opened 2023-04-19
Fix/131

### LINKED ISSUE #131: Gate with a qubit out of range of the Circuit should fail
The following constructs a circuit which cannot be used: the target of the gate does not exist.
julia> QuantumCircuit(qubit_count = 1, gates=[sigma_x(5)]);

fSosaRey:
Closes #131

### DIFF
--- a/src/core/quantum_circuit.jl
+++ b/src/core/quantum_circuit.jl
@@ -19,14 +19,7 @@ q[2]:
 """
 Base.@kwdef struct QuantumCircuit
     qubit_count::Int
-    gates::Vector{AbstractGate} = []
-
-    #Constructor without any gates, for instance: QuantumCircuit(qubit_count=3);
-    function QuantumCircuit(qubit_count::Int,gates::Vector{Any}=[])         
-        @assert qubit_count>0 ("$(:QuantumCircuit) constructor requires qubit_count>0. Received: $qubit_count")
-
-        new(qubit_count,gates)
-    end
+    gates::Vector{AbstractGate} = Vector{AbstractGate}([])
     
     function QuantumCircuit(qubit_count::Int,gates::Vector{<:AbstractGate})    
         @assert qubit_count>0 ("$(:QuantumCircuit) constructor requires qubit_count>0. Received: $qubit_count")
--- a/src/core/quantum_circuit.jl
+++ b/src/core/quantum_circuit.jl
@@ -20,7 +20,26 @@ q[2]:
 Base.@kwdef struct QuantumCircuit
     qubit_count::Int
     gates::Vector{AbstractGate} = []
+
+    #Constructor without any gates, for instance: QuantumCircuit(qubit_count=3);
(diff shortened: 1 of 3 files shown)


src/core/quantum_circuit.jl
```diff
@@ -20,8 +20,27 @@ q[2]:
 Base.@kwdef struct QuantumCircuit
     qubit_count::Int
     gates::Vector{AbstractGate} = []
+
+    #Constructor without any gates, for instance: QuantumCircuit(qubit_count=3);
+    function QuantumCircuit(qubit_count::Int,gates::Vector{Any}=[])
```
plerouxanyon, line 25: Will this not match againt something like

`QuantumCircuit(qubit_count=6, gates=Vector{Any}(["not_a_gate"]))`

?

plerouxanyon (changes requested):

src/core/quantum_circuit.jl
fSosaRey: the only possible instance of `Vector{Any}` is an empty array. Passing `gate=[ "123 "]` or `gate=[1,2,3]` results in `MethodError`. Should I add a test for those cases?

plerouxanyon: Passing the example, I posted results in a `MethodError` on line 28 when it tries to convert a `String` to an `AbstractGate`. It does match the function with a non-empty Any vector.

```julia
QuantumCircuit(qubit_count=6, gates=Vector{Any}(["not_a_gate"]))
```

Results in
```
ERROR: MethodError: Cannot `convert` an object of type String to an object of type AbstractGate
Closest candidates are:
  convert(::Type{T}, ::T) where T at Base.jl:61
Stacktrace:
  [1] setindex!(A::Vector{AbstractGate}, x::String, i1::Int64)
    @ Base ./array.jl:966
  [2] _unsafe_copyto!(dest::Vector{AbstractGate}, doffs::Int64, src::Vector{Any}, soffs::Int64, n::Int64)
    @ Base ./array.jl:253
  [3] unsafe_copyto!
    @ ./array.jl:307 [inlined]
  [4] _copyto_impl!
    @ ./array.jl:331 [inlined]
  [5] copyto!
    @ ./array.jl:317 [inlined]
  [6] copyto!
    @ ./array.jl:343 [inlined]
  [7] copyto_axcheck!
    @ ./abstractarray.jl:1127 [inlined]
  [8] Array
    @ ./array.jl:626 [inlined]
  [9] convert
    @ ./array.jl:617 [inlined]
 [10] QuantumCircuit(qubit_count::Int64, gates::Vector{Any})
    @ Snowflake ~/codes/anyon/github/Snowflake.jl/src/core/quantum_circuit.jl:28
 [11] #QuantumCircuit#27
    @ ./util.jl:493 [inlined]
 [12] top-level scope
    @ REPL[1]:1
```

plerouxanyon (approved):

Issue #148, opened 2023-04-21
Move SnowflakePlots.jl into its own repository


plerouxanyon:
* The documentation should reference the new SnowflakePlots.jl repository.

PR #152, opened 2023-04-22
Docs/update readme

plerouxanyon:
I've updated the README and index page of the Snowflake documentation.

Notable mentions:
* I removed mentions of SnowflakePlots. Once, it has moved to its own repository (#148) we should definitely reference it.
* I've replaced the four typical steps of a quantum application with the three typical steps of submitting a circuit to a quantum service. The four steps did not seem typical at all. (Shor's algorithm, QAQO, VQE are examples which do not follow these steps)
* The centring of the CNOT circuit image does not render on GitHub. I'll still try to fix this.


alireza-najafi:
@plerouxanyon There is no info on transpilation on the readme file. I think we should add one example.

alireza-najafi (approved):

PR #174, opened 2023-04-25
Feat/add reduce swap transpiler

### LINKED ISSUE #124: Add reduce swap transpiler
Before qubit placement, swaps can be implemented by adjusting the qubit indices of gates.

npoirie1:
Adds a transpiler stage that removes Swap gates by moving the gates preceding each Swap gate.

Fixes #124

### DIFF
--- a/src/core/quantum_gate.jl
+++ b/src/core/quantum_gate.jl
@@ -75,7 +75,9 @@ get_connected_qubits(gate::AbstractGate)=[gate.target]
 
 get_gate_parameters(gate::AbstractGate)=Dict()
 
-is_gate_type(gate::AbstractGate, type::Type) = isa(gate, type)
+is_gate_type(gate::AbstractGate, type::Type)::Bool = isa(gate, type)
+
+get_gate_type(gate::AbstractGate)::Type = typeof(gate)
 
 struct MovedGate <:AbstractGate
     original_gate::AbstractGate
@@ -92,7 +94,9 @@ Base.inv(gate::MovedGate) = MovedGate(inv(gate.original_gate), gate.connected_qu
 
 get_gate_parameters(gate::MovedGate) = get_gate_parameters(gate.original_gate)
 
-is_gate_type(gate::MovedGate, type::Type) = isa(gate.original_gate, type)
+is_gate_type(gate::MovedGate, type::Type)::Bool = isa(gate.original_gate, type)
+
+get_gate_type(gate::MovedGate)::Type = typeof(gate.original_gate)
 
 """
     move_gate(gate::AbstractGate,
--- a/src/core/quantum_gate.jl
+++ b/src/core/quantum_gate.jl
@@ -1090,6 +1090,12 @@ Base.inv(gate::RotationX) = rotation_x(gate.target, -gate.theta)
 
 get_gate_parameters(gate::RotationX)=Dict("theta" =>gate.theta)
 
+function move_gate(gate::RotationX,
+    new_connected_qubits::AbstractVector{<:Integer})::AbstractGate
+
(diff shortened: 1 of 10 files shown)


src/core/quantum_gate.jl
```diff
@@ -75,6 +75,106 @@ get_connected_qubits(gate::AbstractGate)=[gate.target]
...
+    .    1.0 + 0.0im
+    1.0 + 0.0im    .
+
+
+```
+"""
+function move_gate(gate::AbstractGate,
+    new_connected_qubits::AbstractVector{T})::AbstractGate where T<:Integer
+
+    return typeof(gate)(new_connected_qubits[1])
```
plerouxanyon, line 175: This default is not a good default and could introduce a lot of very subtle bugs. If a default is provided it should work. It does not have to work optimally, but it should work. There are a lot of cases this will crash if it not specialize (all parameterized gates, all multi-qubit gates). There are also cases where this will produce the wrong output. For instance, if you create a gate with a variable number of qubits, such as a QFT gate, and forget to specialize this method then it won't crash and simply give a wrong answer. Bugs like that will be hard to track down.

It is also a lot of work to specialize this function for every gate that is not an unparameterized single qubit gate. Every-time someone creates a gate then they'll have to add this method as well. We can do this only once by creating a `MovedGate` this gates dispatches to the underlying gate for everything except for `get_connected_qubits`.

src/core/transpile.jl
```diff
@@ -1417,4 +1417,89 @@ function transpile(::CompressRzGatesTranspiler, circuit::QuantumCircuit)::Quantu
...
+    connected_qubits = get_connected_qubits(gate)
+    if haskey(qubit_mapping, connected_qubits[1])
+        outlet_qubit_2 = qubit_mapping[connected_qubits[1]]
+        pop!(qubit_mapping, connected_qubits[1])
+        add_first_swap_branch_mapping!(qubit_mapping, connected_qubits)
+        qubit_mapping[connected_qubits[2]] = outlet_qubit_2
+    else
+        add_first_swap_branch_mapping!(qubit_mapping, connected_qubits)
+        qubit_mapping[connected_qubits[2]] = connected_qubits[1]
+    end
```
plerouxanyon, line 1492: Does it make sense to move `add_first_swap_branch_mapping` to after the if and inline it?

```julia
if haskey(qubit_mapping, connected_qubits[1])
    # ...
else
   # ...
end

if haskey(qubit_mapping, connected_qubits[2])
    # ...
else
   # ...
end
```

plerouxanyon (changes requested):
Thanks for the PR. I've given some comments.

src/core/quantum_gate.jl
npoirie1: Creating a `MovedGate` does remove the issue of having to write a `get_connected_qubits` function for each gate. However, a drawback of using `MovedGate` is that checking the gate type using `isa` returns `MovedGate` instead of the original gate type. This will cause problems in the transpiler. As a solution, we could create a function called `is_gate_type` which makes a type comparison with the original gate and not `MovedGate` directly.

Do you have another solution for this problem?

plerouxanyon: I'm happy with having an `is_gate_type` function. It will also be useful if we are to add multiple gates with different names.

npoirie1: The functions `is_gate_type` and `get_gate_type` have been added. For `AbstractGate` objects, they should be used instead of `isa` and `typeof`.

plerouxanyon (approved):

src/core/transpile.jl
npoirie1: The `update_qubit_mapping!` function has been restructured as you suggested.

PR #183, opened 2023-04-26
Docs/asynchronous jobs

### LINKED ISSUE #163: Add documentation on running jobs assynchronously using tasks

### LINKED ISSUE #180: README.md uses CNOT image but should be Bell Circuit image
README.md contains the following snippet under the "Getting Started" section

```
The quantum circuit for generating a Bell state involves a Hadamard gate on one of the qubits followed by a CNOT gate (see [here](https://en.wikipedia.org/wiki/Quantum_logic_gate) for an introduction to quantum logic gates). This circuit is shown below:

<div style="text-align: center;">
	<img
		src="./docs/src/images/cnot_circuit.svg"
		title="Bell state generator circuit"
		width="240"
	/>
</div>

```
Note that the `img`'s src references the `images/cnot_circuit.svg` file and that image is just a CNOT circuit ... it does not include a Hadamard gate.

We should create and use something like `bell_circuit.svg` with the appropriate gates shown.

plerouxanyon:
Closes #163 
Closes #180


examples/asynchronous_jobs.jl
```diff
@@ -0,0 +1,32 @@
...
+])
+
+user = ENV["ANYON_QUANTUM_USER"]
+token = ENV["ANYON_QUANTUM_TOKEN"]
+host = ENV["ANYON_QUANTUM_HOST"]
+
+client = Client(host=host, user=user, access_token=token)
+qpu = AnyonQPU(client)
+
+num_repititions = 200
```
fSosaRey, line 15: num_repetitions

fSosaRey (changes requested):

docs/src/examples/advanced/asynchronous_jobs.md
```diff
@@ -0,0 +1,104 @@
...
+
+```
+
+We must remember to schedule the task; otherwise, Julia will not know that it should start running your task! Next, we need to yield execution of the current thread to the newly scheduled task to ensure that the scheduler starts with the task. Otherwise, the task will be scheduled, but it might not submit a job to the quantum computer any time soon! After yielding once, we can continue to do work before we [fetch](https://docs.julialang.org/en/v1/base/parallel/#Base.fetch-Tuple{task}) the results from that task.
+
+
+```jldoctest asynchronous_job; output = false
+yieldto(task)
+
+# Simulate work by calculating the nth Fibonacci number slowly
```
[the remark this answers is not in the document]
plerouxanyon: Instead of memoizing the results it re-computes it everytime.

plerouxanyon:
@fSosaRey All the comments have been addressed. Would you have another look?

docs/src/examples/advanced/asynchronous_jobs.md
```diff
@@ -0,0 +1,106 @@
...
+end
+```
+
+This example is going to how to run jobs asynchronously using Julia tasks. Asynchronous jobs allow users to continue other work while waiting for the quantum resources.
+
+## Julia tasks
+
+Communication with a quantum computer happens over a network. Any requests sent to a quantum computer are sent over the network. After the request is sent, we can only continue the job execution when we get a response from the quantum computer.
+
+If we absolutely have to have the job's result, we have no choice but to wait. Often, other work can be done while waiting for the quantum job to complete. We want to be able to suspend the `run_job` so that we can continue with other work until we are ready to use the result. This type of control flow is called asynchronous control flow. We recommend you read [Julia's page on asynchronous programming](https://docs.julialang.org/en/v1/manual/asynchronous-programming/) if you are unfamiliar with it.
```
fSosaRey, line 17: By "suspend" I understand "pause". We want to be able to decouple the execution of `run_job` from the rest of our program?

fSosaRey (approved):

PR #186, opened 2023-04-27
Docs/clean up coverage and move contributing page

### LINKED ISSUE #150: Coverage from `coverage.jl` and CodeCov does not match

plerouxanyon:
Closes #150

### DIFF
--- a/docs/src/development.md
+++ b/docs/src/development.md
@@ -1,0 +1,20 @@
+# Snowflake development
+
+!!! warning
+    This page is under construction 🚧
+
+## Run coverage locally
+
+You run coverage locally from the project directory using
+
+```bash
+julia --project=. coverage.jl
+```
+
+The script returns the covered and total line as output. An example output is shown below
+
+```text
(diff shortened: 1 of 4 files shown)


docs/src/development.md
```diff
@@ -0,0 +1,20 @@
+# Snowflake development
+
+!!! warning
+    This page is under construction 🚧
+
+## Run coverage locally
+
```
caffeine-storm, line 7: I had to run Pkg.Instantiate() to get the coverage report to run... might worth mentioning something like
```suggestion
If you haven't already, instantiate the project with Julia's package manager.
```bash
julia --project=. -e 'using Pkg; Pkg.Instantiate()'
``` .
```
But please rewrite to more canonical Julia-isms if you can :sweat_smile:

caffeine-storm (approved):
Looks good to me!

There is an optional step we could add to make running coverage a bit more explicit... I'll leave it to your judgement as to whether it's worth including or not.

PR #190, opened 2023-04-28
Feat/sparse operator

alireza-najafi:


src/core/qobj.jl
```diff
@@ -148,6 +150,47 @@ end
...
+"""
+
+struct SparseOperator{T<:Complex}<:AbstractOperator
+    data::  SparseMatrixCSC{T, Int64}
+end
+# Constructor from Real-valued Matrix
+SparseOperator(x::Matrix{T}) where {T<:Real} = SparseOperator(SparseArrays.sparse(Complex.(x)))
+# Constructor from Complex-valued Matrix
+SparseOperator(x::Matrix{T}) where {T<:Complex} = SparseOperator(SparseArrays.sparse(x))
+
```
fSosaRey, line 173: missing a get_matrix(s::SparseOperator), which is used when casting to DenseOperator
```
get_matrix(op::SparseOperator{T}) where {T<:Complex} = convert(Matrix{T},op.data)
```

test/sparse_operator.jl
```diff
@@ -0,0 +1,37 @@
...
+    @test adjoint(A) ≈ A_dag
+
+    @test expected_value(σ_z,spin_up()) ≈ Complex(1.0)
+    @test expected_value(σ_z,spin_down()) ≈ Complex(-1.0)
+
+    @test kron(σ_x,σ_y) ≈ sparse(kron(sigma_x(),sigma_y()))
+
+    @test is_hermitian(σ_x)
+
+    println(σ_x)  #test "show" function
```
fSosaRey, line 36: this gives the following error on my machine:  
```
julia> include("test/sparse_operator.jl")
(2, 2)-element Snowflake.SparseOperator:
Underlying data ComplexF64:
SparseOperator: Error During Test at /home/facundo/codes/anyon/Snowflake.jl/test/sparse_operator.jl:4
  Got exception outside of a @test
  MethodError: no method matching _show_with_braille_patterns(::Base.TTY, ::SparseArrays.SparseMatrixCSC{ComplexF64, Int64})
  Closest candidates are:
    _show_with_braille_patterns(::IOContext, ::SparseArrays.AbstractSparseMatrixCSC) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/SparseArrays/src/sparsematrix.jl:236

```

src/core/qobj.jl
fSosaRey, line 173: missing 
```
apply_operator!(::Ket, ::SparseOperator)
```

test/sparse_operator.jl
```diff
@@ -0,0 +1,37 @@
+using Snowflake
+using Test
+
```
fSosaRey, line 3: add standardized test for all AbstractOperators:

```
include("test_functions.jl")

test_operator_implementation(SparseOperator,dim=2,label="SparseOperator",values=[1.,2.,0.,3.])
 
```

fSosaRey (changes requested):

src/core/qobj.jl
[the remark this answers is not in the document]
alireza-najafi: Why?

test/sparse_operator.jl
alireza-najafi: We will do that in another pull request. Let's start a new issue to address this.

Project.toml
```diff
@@ -4,11 +4,13 @@ authors = ["Alireza Yazdi <alireza@anyonsys.com>", "Paul Le Roux <pleroux@anyons
 version = "0.1.0"
 
 [deps]
-Coverage = "a2441757-f6aa-5fb2-8edb-039e3f45d037"
```
fSosaRey, line 7: Coverage is required to run coverage.jl

src/core/quantum_gate.jl
```diff
@@ -154,6 +154,13 @@ function apply_operator!(
     end
 end
 
+function apply_operator!(
+    state::Ket,
+    operator::SparseOperator,
+    connected_qubit::Vector{<:Integer})
+    @warn "apply_operator! is not implemented for SparseOperator. Try using DenseOperator instead."
```
fSosaRey, line 161: @pleroux0  are we ok with pushing this to main?

fSosaRey (approved):

src/core/quantum_gate.jl
plerouxanyon: This code is dangerous. A warning is easy to miss, and it will lead to incorrect unexpected results. Rather remove it and let it crash or throw an error with a detailed message.

plerouxanyon (changes requested):

Project.toml
plerouxanyon: I think we can get the best of both if we move the `Coverage` dependency to `test/Project.toml`. Snowflake then has no unneeded dependencies, and we can run the `coverage.jl` script from the test project.

test/test_functions.jl
```diff
@@ -150,10 +150,10 @@ function test_operator_implementation(
         end
 
         op=op_type(make_array(dim, ComplexF64, values))
-    
         Snowflake.apply_operator!(ψ_1,op,[v for v in 1:get_num_qubits(op)])
-
-        @test op*ψ_0 ≈ ψ_1
+        if (op_type!=SparseOperator)
```
plerouxanyon, line 154: This doesn't seem right. This is making the functions do different things for different operators (more than one thing) Either make the function generic enough so that this escape hatch is not necessary or split the function in two.

plerouxanyon (changes requested):

plerouxanyon (approved):

PR #198, opened 2023-04-28
fix: allow multi-target multi-parameter gates to be transpiled and printed in circuits

### LINKED ISSUE #193: SwapQubitsForLineConnectivityTranspiler assumes specific constructor of two-qubit gates
If a user would define the following valid and existing gate

```julia
struct ISwapPowGate <: AbstractGate
        targets::Vector{Int}
        theta::Real
    end

get_connected_qubits(gate::ISwapPowGate)=gate.targets
get_gate_parameters(gate::ISwapPowGate)=Dict("theta"=>gate.theta)
get_operator(gate::ISwapPowGate, T::Type{<:Complex}=ComplexF64) = DenseOperator(T.[
    [1,0,0,0],
    [0, cos(gate.theta/2), im*sin(gate.theta/2), 0],
    [0, im*sin(gate.theta/2), im*cos(gate.theta/2), 0],
    [0,0,0,1],
])
```

The `SwapQubitsForLineConnectivityTranspiler` would fail.

fSosaRey:
Closes #193

### DIFF
--- a/src/core/transpile.jl
+++ b/src/core/transpile.jl
@@ -1239,13 +1239,15 @@ function transpile(::SwapQubitsForLineConnectivityTranspiler, circuit::QuantumCi
     output_circuit=QuantumCircuit(qubit_count=qubit_count)
 
     for gate in gates
-
+        
         connected_qubits=get_connected_qubits(gate)
         (consecutive_mapping,sorting_order)=remap_qubits_to_consecutive(connected_qubits)
+        
+        params=get_gate_parameters(gate)
 
         if length(consecutive_mapping)>1
     
-            gates_block=[get_gate_type(gate)(consecutive_mapping...)]
+            gates_block=[get_gate_type(gate)(consecutive_mapping...,values(params)...)]
 
             @assert get_connected_qubits(gates_block[1])==consecutive_mapping (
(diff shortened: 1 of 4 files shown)


fSosaRey:
coverage is not full because no multi-target parameterized is present in the library

src/core/transpile.jl
```diff
@@ -1239,13 +1239,15 @@ function transpile(::SwapQubitsForLineConnectivityTranspiler, circuit::QuantumCi
...
+        
         connected_qubits=get_connected_qubits(gate)
         (consecutive_mapping,sorting_order)=remap_qubits_to_consecutive(connected_qubits)
+        
+        params=get_gate_parameters(gate)
 
         if length(consecutive_mapping)>1
     
-            gates_block=[get_gate_type(gate)(consecutive_mapping...)]
+            gates_block=[get_gate_type(gate)(consecutive_mapping...,values(params)...)]
```
plerouxanyon, line 1250: As discussed in Teams, this forces gates to have specific constructors. We can use `move_gate` and `MovedGate` as a more flexible solution.

plerouxanyon (changes requested):

PR #202, opened 2023-05-01
Refactor/use slurping for push

### LINKED ISSUE #200: Use slurping to push gates to circuits
The function which pushes `AbstractGate` objects into `QuantumCircuit` objects should use slurping like the `push!` function in `Base`. See [https://docs.julialang.org/en/v1/base/collections/#Base.push!](https://docs.julialang.org/en/v1/base/collections/#Base.push!).

Passing a vector of `AbstractGate` objects would require splatting. For more details about slurping and splatting, see https://docs.julialang.org/en/v1/manual/faq/#The-two-uses-of-the-...-operator:-slurping-and-splatting.

npoirie1:
Enables slurping for pushing `AbstractGate` objects to a `QuantumCircuit`. The function `push!(circuit::QuantumCircuit, gates::AbstractGate...)` now better matches `Base.push!`.

More details about slurping are provided here: https://docs.julialang.org/en/v1/manual/faq/#What-does-the-...-operator-do?

Fixes #200

### DIFF
--- a/docs/src/getting_started.md
+++ b/docs/src/getting_started.md
@@ -47,8 +47,8 @@ q[2]:
 In Snowflake, all qubits start in state $\left|0\right\rangle$. Our circuit is, therefore,  in state $\left|00\right\rangle$. We now proceed by adding gates to our circuit.
 
 ```jldoctest getting_started
-push!(c, [hadamard(1)])
-push!(c, [control_x(1, 2)])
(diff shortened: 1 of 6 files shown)


npoirie1:
SnowflakePlots will also need to be updated to reflect these changes. I will add an issue in The SnowflakePlots repository once the branch is merged.

Issue #100, opened 2023-03-23
Add optimized controlled operator


plerouxanyon:

plerouxanyon:
None of the gates that Snowflake will include will use this operator. This operator is, therefore, not essential for the 0.1 release, but it is still a nice to have.

PR #205, opened 2023-05-01
feat: AbstractControlledGate interface

### LINKED ISSUE #77: Add general controlled operator

### LINKED ISSUE #78: Optimized CNOT operator

### LINKED ISSUE #98: Optimize Toffoli gate

### LINKED ISSUE #105: Optimize CZ gate

fSosaRey:
Closes #77 
Closes #78 
Closes #98 
Closes #105


fSosaRey:
Closes #77 
Closes #78

src/core/quantum_gate.jl
```diff
@@ -166,6 +166,34 @@ end
 
 get_connected_qubits(gate::MovedGate) = gate.connected_qubits
 
+function get_control_qubits(gate::MovedGate)::Vector{Int}
```
plerouxanyon, line 169: We actually need a `MovedControlledGate` since a `MovedGate` is not a subtype of `ControlledGate`.

plerouxanyon (changes requested):
Something that is missing, but I think we should add in a future PR is a `ControlledOperator` with an optimized `apply_operator!`

plerouxanyon:
I realize there already is an issue for adding a controlled operator (#100)

fSosaRey:
> I realize there already is an issue for adding a controlled operator (#100)

Adding a ControlledOperator<:AbstractOperator would mean defining an object that is identical to DenseOperator, except at dispatch for apply_operator. There would be many additions in the whole library for little added value. We could instead call apply_controlled_operator, for the handful of gates which are controlled, i.e., ControlX, ControlZ and Toffoli.

plerouxanyon:
> Adding a ControlledOperator<:AbstractOperator would mean defining an object that is identical to DenseOperator, except at dispatch for apply_operator. There would be many additions in the whole library for little added value. We could instead call apply_controlled_operator, for the handful of gates which are controlled, i.e., ControlX, ControlZ and Toffoli.

I don't think that is true. `ControlledOperators` would only need to store the gate that will be applied to target and the number of control qubits.

```julia
struct ControlledOperator <: AbstractOperator
    num_control_qubits::Int
    target::AbstractOperator
end
```

Unless, I am making a mistake, it doesn't look like a dense operator to me at all.

src/core/quantum_gate.jl
```diff
@@ -1464,6 +1754,38 @@ get_operator(gate::Toffoli, T::Type{<:Complex}=ComplexF64) = toffoli(T)
 
 get_connected_qubits(gate::Toffoli)=[gate.control_1, gate.control_2, gate.target]
 
+get_control_qubits(gate::Toffoli)=[gate.control_1, gate.control_2]
+
+get_target_qubits(gate::Toffoli)=[gate.target]
+
+# optimized application of ControlX, ControlZ or Toffoli gate without calling operator 
+# (it is hard-coded in apply_control_x!, apply_control_z! or apply_toffoli!, respectively)
+function apply_gate!(state::Ket, gate::Union{ControlX,ControlZ,Toffoli})
```
plerouxanyon, line 1763: This doesn't look great since this is essentially three functions and we doing a union to share code which looks like a code smell. What we are in fact repeating is the validation step which we have in every `apply_gate`.

We can refactor (but in a later PR) `apply_gate` into `apply_gate` into `unsafe_apply_gate` where `apply_gate` does the validation and then passes the values onto the `unsafe_apply_gate` function. All the shared validation then happens once. (All gates are required to be inside the circuit) and it dispatches to `unsafe_apply_gate` which can then handle the gate specifics. Then we don't have to have Union constructs and then by-hand dispatch to the correct function.

EDIT:

I've created an issue for the refactor so that we can do it later: #214.

plerouxanyon (changes requested):
Almost there!

src/core/quantum_gate.jl
```diff
@@ -229,7 +267,11 @@ function move_gate(gate::AbstractGate,
         end
     end
     if found_move
-        return MovedGate(gate, new_connected_qubits)
+        if gate isa AbstractControlledGate
```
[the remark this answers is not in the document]
fSosaRey: having a `MovedGate(gate::AbstractControlledGate)=MovedControlledGate(gate)` avoids duplicating the move_gate() function

plerouxanyon (approved):

PR #231, opened 2023-05-17
Add link checking script for Markdown files and fix a 404 it found

caffeine-storm:


caffeine-storm:
I would add a call to `docs/test/check-for-404.sh` to Github CI but, until all the 404s are fixed, I imagine it would block MRs? Let me know if we should add it now or if there's a way for it to warn us without blocking merges.

plerouxanyon (approved):

Issue #226, opened 2023-05-08
Refactor AbstractGate to split placement and gate information


plerouxanyon:
# Overview

We want to split the `AbstractGate` interface into two parts. The placement information and symbol information.

## Example

```julia
struct PlacedGate begin
    gate::AbstractGate
    connected_qubits::Vector{Int}
end

struct H <: AbstractGate begin
end

get_operator(gate::H)::AbstractOperator = DenseOperator([...])
get_num_connected_qubits(gate::H)::Int = 1

struct Rz <: AbstractGate begin
    phase::Real
end
```

## Justification

We currently group both placement information and the gate symbol into the `AbstractGate` interface. The only way to move a qubit given only the interface was to define a `MovedGate`, which embeds the original gate and exposes the interface, but with different connectivity. This seemed to work well. Then @npoirie1 pointed out that the wrapper type interferes with Julia's dispatch rules. We found missing test cases where `MovedGate`s were not tested and caused bugs when passed to some transpilers. To keep with the current architecture, all transpilers (the transpilers in Snowflake and those that are going to be defined by users), would have to handle moved gates correctly. This adds more work every time a new transpiler is written. An alternative is to make `move_gate` part of the `AbstractGate` interface, but then it adds work every time a new gate is written. The root issue was that we coupled the placement information and the gate symbol by making them part of the same interface. The long term solution is to split the interface.

See #224 for more information.

PR #232, opened 2023-05-18
Feat/gate placement

plerouxanyon:
This is the first change required for #226.

An intermediate type, `GatePlacement`, is added as a drop-in replacement for the `AbstractGate` type. It replaces the use of `AbstractGate` in `QuantumCircuit`. This means that all code downstream of  `QuantumCircuit` now interacts with `GatePlacement` instead of `AbstractGate`, 

There are two parts remaining.

1. The `GatePlacement` interface needs to be trimmed to not include the gate interface. (and all downstream code of `QuantumCircuit` needs to be updated accordingly)
2. We can remove the connection information from the `AbstractGate` interface (and all the gates).


caffeine-storm (changes requested):

plerouxanyon:
We are busy with talks about if this is indeed the way forward. I marked the PR as a draft to signify that even though all the comments will soon be addressed, the PR will not be merged as soon as possible.

src/core/quantum_gate.jl
```diff
@@ -101,6 +107,119 @@ get_connected_qubits(gate::AbstractGate)=[gate.target]
 
 get_gate_parameters(gate::AbstractGate)=Dict{String,Real}()
 
+get_num_connected_qubits(gate::AbstractGate)=length(get_connected_qubits(gate))
+
+"""
+    GatePlacement
```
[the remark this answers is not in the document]
plerouxanyon: Renamed to `PlacedGate`

caffeine-storm (approved):
LGTM!

plerouxanyon:
I've rebased to the latest changes in main, and I've done the rename (AbstractGate -> AbstractGateSymbol, GatePlacement -> Gate).

@alireza-najafi @caffeine-storm The code is ready for another look.

src/core/quantum_gate.jl
```diff
@@ -93,21 +99,134 @@ q[2]:───────
...
+
+Connected_qubits	: [3]
+Operator:
+(2, 2)-element Snowflake.DenseOperator:
+Underlying data ComplexF64:
+0.7071067811865476 + 0.0im    -4.329780281177466e-17 - 0.7071067811865475im
+4.329780281177466e-17 - 0.7071067811865475im    0.7071067811865476 + 0.0im
+```
+"""
+struct Gate <: AbstractGateSymbol
```
fSosaRey, line 149: is the composition pattern here temporary? Seems clumsy to have Gate be a subtype of AbstractGateSymbol, meaning some AbstractGateSymbol have placement information and some don't. Also, if we use a generic struct such as 
```
struct Gate{T} where T<:AbstractGateSymbol
    gate::T
    connected_qubits::Vector{Int}
    (...)
end
```
, this would enable constraining outputs of `compress_to_universal` for instance to `Gate{Universal}`, whereas it now is the all-encompassing `Gate`.

src/core/quantum_gate.jl
```diff
@@ -305,7 +417,7 @@ julia> print(ψ_0)
 
 ```
 """
-function apply_gate!(state::Ket, gate::AbstractGate)
+function apply_gate!(state::Ket, gate::AbstractGateSymbol)
```
fSosaRey, line 420: this only works for `gate::Gate`, not if gate is a raw `AbstractGateSymbol` with no placement info

src/core/quantum_gate.jl
```diff
@@ -322,6 +434,9 @@ function apply_gate!(state::Ket, gate::AbstractGate)
     apply_operator!(state,operator,connected_qubits)
 end
 
+# TODO(#226): delete on completion
+apply_gate!(state::Ket, gate::Gate) = apply_gate!(state::Ket, get_gate(gate))
```
fSosaRey, line 438: It seems to me this is the signature we should keep

src/core/transpile.jl
```diff
@@ -86,7 +86,7 @@ end
 struct CompressSingleQubitGatesTranspiler<:Transpiler end
 
 # convert a single-target gate to a Universal gate
-function as_universal_gate(target::Integer,op::AbstractOperator)::Universal
+function as_universal_gate(target::Integer,op::AbstractOperator)::Gate
```
fSosaRey, line 89: as mentioned above, changing Gate to a generic struct could better constrain output type here

src/core/transpile.jl
```diff
@@ -115,24 +115,20 @@ function as_universal_gate(target::Integer,op::AbstractOperator)::Universal
 end
 
 # compress (combine) several single-target gates with a common target to a Universal gate
-function compress_to_universal(gates::Vector{<:AbstractGate})::Universal
+function compress_to_universal(gates::Vector{<: AbstractGateSymbol}, target::Int)::Gate
```
fSosaRey, line 118: Input of `gates::Vector{<: AbstractGateSymbol}` would fail if the elements are not (placed) Gate types.

src/core/transpile.jl
```diff
@@ -254,7 +250,7 @@ function find_and_compress_blocks(
                 gates_block=[gates[i] for i in block]
 
                 if length(block)>1
-                    push!(output_circuit,compression_function(gates_block))
+                    push!(output_circuit, compression_function(gates_block, target))
```
fSosaRey, line 253: the logic in `find_and_compress_blocks` already groups candidates by common-target. We could have an unsafe_compression_function which assumes common-target, and only interact with inner (AbstractGateSymbol) objects, not placed Gates

src/core/transpile.jl
```diff
@@ -699,16 +691,14 @@ function simplify_rx_gate(
...
     params=get_gate_parameters(gate)
    
-    target=get_connected_qubits(gate)[1]
-
     theta   =params["theta"]
     phi     =params["phi"]
     lambda  =params["lambda"]
 
-    gate_array=Vector{AbstractGate}([])
+    gate_array=Vector{AbstractGateSymbol}([])
```
fSosaRey, line 701: `gate_array=Vector{Gate}([]) ` would enphasize that these are placed Gates

src/core/transpile.jl
```diff
@@ -848,16 +840,14 @@ end
...
     params=get_gate_parameters(gate)
    
-    target=get_connected_qubits(gate)[1]
-
     γ=params["theta"]
     β=params["phi"]+π/2
     δ=params["lambda"]-π/2
    
-    gate_array=Vector{AbstractGate}([])
+    gate_array=Vector{AbstractGateSymbol}([])
```
fSosaRey, line 850: same comment as above

src/core/transpile.jl
```diff
@@ -931,33 +921,35 @@ function transpile(::CastUniversalToRzRxRzTranspiler, circuit::QuantumCircuit)::
...
-function cast_rx_to_rz_and_half_rotation_x(gate::RotationX)::Vector{AbstractGate}
+function cast_rx_to_rz_and_half_rotation_x(gate::RotationX)::Vector{AbstractGateSymbol}
     params=get_gate_parameters(gate)
    
     target=get_connected_qubits(gate)[1]
 
     theta   =params["theta"]
 
-    gate_array=Vector{AbstractGate}([])
+    gate_array=Vector{AbstractGateSymbol}([])
```
fSosaRey, line 952: and here

src/core/transpile.jl
```diff
@@ -1397,7 +1389,7 @@ end
 struct CompressRzGatesTranspiler<:Transpiler end
 
 # construct a PhaseShift gate from an input Operator
-function as_phase_shift_gate(target::Integer,op::AbstractOperator)::PhaseShift
+function as_phase_shift_gate(target::Integer,op::AbstractOperator)::Gate
```
fSosaRey, line 1392: could be constrained to Gate{PhaseShift} by modifying Gate struct to be generic

src/core/transpile.jl
```diff
@@ -1418,24 +1410,15 @@ function as_phase_shift_gate(target::Integer,op::AbstractOperator)::PhaseShift
 end
 
 # compress (combine) several Rz-type gates with a common target to a PhaseShift gate
-function compress_to_rz(gates::Vector{<:AbstractGate})::PhaseShift
+function compress_to_rz(gates::Vector{<: AbstractGateSymbol}, target::Int)::Gate
```
fSosaRey, line 1413: same comment

fSosaRey (changes requested):

src/core/quantum_gate.jl
```diff
@@ -1,13 +1,13 @@
...
+A `Gate` is an instantiation of an `AbstractGateSymbol`, which can be added to a `QuantumCircuit` in order to apply an operator to one or more `target` qubits.
+`AbstractGateSymbol` is useful to dispatch all `Gates` to default implementation of functions such as get_connected_qubits(). 
 Those functions are then specialized for `Gates` requiring a different implementation. 
 
-`AbstractGate` is an abstract type, which means that it cannot be instantiated. 
-Instead, each concrete type of `Gate` is a struct which is a subtype of `AbstractGate`.
-Each descendant of `AbstractGate` must have at least the following fields:
+`AbstractGateSymbol` is an abstract type, which means that it cannot be instantiated. 
+Instead, each concrete type of `Gate` is a struct which is a subtype of `AbstractGateSymbol`.
+Each descendant of `AbstractGateSymbol` must have at least the following fields:
```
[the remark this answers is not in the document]
plerouxanyon: I've updated the documentation since neither of those fields is required, and only the `get_operator` and `get_num_connected_qubits` functions are required.

src/core/quantum_gate.jl
```diff
@@ -93,21 +99,134 @@ q[2]:───────
...
+Equivalent DenseOperator:
+1.0 + 0.0im    0.0 + 0.0im    0.0 + 0.0im    0.0 + 0.0im
+0.0 + 0.0im    0.0 + 0.0im    0.0 + 1.0im    0.0 + 0.0im
+0.0 + 0.0im    0.0 + 1.0im    0.0 + 0.0im    0.0 + 0.0im
+0.0 + 0.0im    0.0 + 0.0im    0.0 + 0.0im    1.0 + 0.0im
+```
+
+
+```jldoctest
+julia> gate_placement = Gate(universal(1, pi/2, -pi/2, pi/2), [3])
```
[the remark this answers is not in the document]
plerouxanyon: Yes, it will be removed, but that will be in a future PR. I updated the targets to be consistent and renamed `gate_placement` to `gate`.

src/core/quantum_gate.jl
```diff
@@ -93,21 +99,134 @@ q[2]:───────
...
+
+
+"""
+    is_gate_type(gate::AbstractGateSymbol, type::Type)::Bool 
 
 Determines if a `gate` is of the specified `type`.
 
 !!! warning "Use is_gate_type instead of isa!"
-    For `AbstractGate` objects, `is_gate_type` should be used instead of `isa`. The
+    For `AbstractGateSymbol` objects, `is_gate_type` should be used instead of `isa`. The
```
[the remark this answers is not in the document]
plerouxanyon: I've added the missing TODO comments.

src/core/quantum_gate.jl
plerouxanyon: Since this is two points

1. Gate should not be inheriting from AbstractGateSymbol
2. Making gates universal makes sense

I am going to answer them separately.

plerouxanyon: 1. Yes, it is a stop-gap measure to allow partial implementation of #226. The inheritance will be removed once #226 is complete. I've added a TODO comment to indicate this.

plerouxanyon: 2. I agree that it is much cleaner. I am going to start implementing that.

src/core/quantum_gate.jl
plerouxanyon: See above.

src/core/quantum_gate.jl
plerouxanyon: 2. Gate is now parametric

src/core/transpile.jl
plerouxanyon: Updated

src/core/transpile.jl
plerouxanyon: I agree, but let's handle this in a future PR. I've made an issue:  https://github.com/anyonlabs/Snowflake.jl/issues/249.

src/core/transpile.jl
plerouxanyon: Updated

src/core/transpile.jl
plerouxanyon: Updated

src/core/transpile.jl
plerouxanyon: Updated

src/core/transpile.jl
plerouxanyon: Updated

src/core/transpile.jl
plerouxanyon: Updated

src/core/quantum_gate.jl
```diff
@@ -93,21 +99,143 @@ q[2]:───────
...
+Connected_qubits	: [3]
+Operator:
+(2, 2)-element Snowflake.DenseOperator:
+Underlying data ComplexF64:
+0.7071067811865476 + 0.0im    -4.329780281177466e-17 - 0.7071067811865475im
+4.329780281177466e-17 - 0.7071067811865475im    0.7071067811865476 + 0.0im
+```
+"""
+struct Gate{GateType} <: AbstractGateSymbol where GateType <: AbstractGateSymbol
+    gate::GateType
```
fSosaRey, line 152: changing the `gate` field to `gate_symbol` would be less ambiguous. Then `get_gate(gate::Gate)` could be `get_gate_symbol(gate::Gate)

src/core/quantum_gate.jl
```diff
@@ -93,21 +99,143 @@ q[2]:───────
...
+    parameters = get_gate_parameters(gate)
+
+    if isempty(parameters)
+        show_gate(io, get_gate_type(gate), targets, get_operator(gate))
+    else
+        show_gate(io, get_gate_type(gate), targets, get_operator(gate), parameters)
+    end
+end
+
+function get_gate(gate::Gate)::AbstractGateSymbol
```
fSosaRey, line 184: As mentioned above, the signature `get_gate(gate)` is disconcerting. How about `get_gate_symbol(gate)` ?

src/core/transpile.jl
```diff
@@ -115,24 +115,20 @@ function as_universal_gate(target::Integer,op::AbstractOperator)::Universal
 end
 
 # compress (combine) several single-target gates with a common target to a Universal gate
-function compress_to_universal(gates::Vector{<:AbstractGate})::Universal
+function compress_to_universal(gates::Vector{Gate{AbstractGateSymbol}}, target::Int)::Gate{Universal}
```
fSosaRey, line 118: the type signature `gates::Vector{Gate}` would be sufficient here

src/core/transpile.jl
```diff
@@ -251,10 +247,10 @@ function find_and_compress_blocks(
 
                 block=blocks_per_target[target][block_index]
 
-                gates_block=[gates[i] for i in block]
+                gates_block::Vector{Gate{AbstractGateSymbol}} = [ convert(Gate, gates[i]) for i in block]
```
fSosaRey, line 250: same comment, `::Vector{Gate}` is less off-putting

test/transpile.jl
```diff
@@ -573,10 +573,10 @@ end
 end
 
 @testset "cast_to_cz: unkown gate" begin
```
fSosaRey, line 575: typo: unknown

fSosaRey (approved):
minor cosmetic changes, otherwise, looking good!

plerouxanyon:
I've updated this PR with the latest changes in master. I didn't fix the coverage yet. I'll do that next.

fSosaRey (approved):

src/core/quantum_gate.jl
```diff
@@ -1744,7 +2024,7 @@ get_operator(gate::PhaseShift,T::Type{<:Complex}=ComplexF64) = phase_shift(gate.
 
 Base.inv(gate::PhaseShift) = phase_shift(gate.target, -gate.phi)
 
-get_gate_parameters(gate::PhaseShift)=Dict("phi" =>gate.phi)
+get_gate_parameters(gate::PhaseShift)=Dict("lambda" =>gate.phi)
```
fSosaRey: inconsistent key name and internal field, why did we change it?

fSosaRey (approved):

src/core/quantum_circuit.jl
```diff
@@ -24,12 +24,10 @@ Base.@kwdef struct QuantumCircuit
     function QuantumCircuit(qubit_count::Int,gates::Vector{GateType}) where GateType <: Gate
         @assert qubit_count>0 ("$(:QuantumCircuit) constructor requires qubit_count>0. Received: $qubit_count")
     
-        c=new(qubit_count,[])
-
-        # add gates, with ensure_gate_is_in_circuit()
-        append!(c,gates)
```
fSosaRey, line 30: I have no objection to the change, but I dont see why this no longer works

src/core/quantum_gate.jl
fSosaRey: @plerouxanyon we should rename the internal `gate.phi `to `gate.lambda`

src/core/quantum_circuit.jl
plerouxanyon: It failed precompilation using Julia 1.9 when calling `Pkg.test()`. It might have been a bug in the precompile tools, but since the fix was so simple, I didn't spend time investigating further.

src/core/quantum_gate.jl
plerouxanyon: We changed it in https://github.com/anyonlabs/Snowflake.jl/pull/261.  Our quantum computer expects an "rz" symbol with a "lambda" parameter. The problem is that we are coupling the gate definition with the textual symbol that we send to Anyon's quantum computer. We should be able to use `PhaseShift` with a `phi` parameter and convert that to `rz` with a `lambda` parameter only when we use `AnyonQPU`. I will make a follow-up issue for this.

plerouxanyon: I've created #282 to address the issue.

PR #244, opened 2023-06-05
ControlledGate implementation

### LINKED ISSUE #236: Add support for arbitrary controlled gates
We want to be able to generate a [W-state](https://en.wikipedia.org/wiki/W_state), and for that, we need a controlled-Hadamard gate.

Since a Hadamard gate is dense, it makes sense to go ahead and make a generic controlled gate class.

fSosaRey:
Kernel is defined using DenseOperator{2}, and apply_gate dispatches to specialized apply_controlled_gate!

We could add a specialization using a Diagonal kernel or an Antidiagonal kernel, but the Pauli gates fitting those descriptions are already optimized individually.

DenseOperators of arbitraty size, (multiple controls) could also be created, but again, Toffoli has its optimization. 

Closes #236

### DIFF
--- a/src/core/quantum_gate.jl
+++ b/src/core/quantum_gate.jl
@@ -842,7 +842,7 @@ apply_controlled_gate_operator!(
 
 apply_controlled_gate_operator!(
     state::Ket,
-    full_operator::DenseOperator{4},
+    full_operator::DenseOperator,
     kernel::AbstractOperator,
     connected_qubits::Vector{Int}
     ) = throw(NotImplementedError(:apply_controlled_gate_operator!,kernel))
--- a/src/core/quantum_gate.jl
+++ b/src/core/quantum_gate.jl
@@ -103,7 +103,7 @@ In addition, a `ControlledGate{X45}` can be constructed using:
 ```jldoctest gate_struct
 julia> control=1; target=2;
 
-julia> ControlledGate(x_45,control, target)
+julia> ControlledGate(x_45(target),control)
 Gate Object: ControlledGate{X45}
 Connected_qubits	: [1, 2]
 Operator:
@@ -143,7 +143,7 @@ controlled Hadamard gate is constructed using the `hadamard()` `Function`. For i
 of a `ControlledGate{Hadamard}` with `control=1` and `target=2` is performed by calling:
 
 ```jldoctest controlled_hadamard
-julia> controlled_hadamard=ControlledGate(hadamard,1,2)
+julia> controlled_hadamard=ControlledGate(hadamard(2),1)
 Gate Object: ControlledGate{Snowflake.Hadamard}
 Connected_qubits	: [1, 2]
 Operator:
@@ -174,7 +174,7 @@ constructed. For instance, the following constructs the equivalent of a `Toffoli
 but as a `ConnectedGate{SigmaX}`, with `control_qubits=[1,2]` and `target_qubit=[3]`:
 
 ```jldoctest 
-julia> toffoli_as_controlled_gate=ControlledGate(sigma_x,[1,2],[3])
+julia> toffoli_as_controlled_gate=ControlledGate(sigma_x(3),[1,2])
 Gate Object: ControlledGate{Snowflake.SigmaX}
 Connected_qubits	: [1, 2, 3]
 Operator:
@@ -199,11 +199,13 @@ struct ControlledGate{GateType<:AbstractGate}<:AbstractGate
     kernel_qubits::Vector{Int}
 
     function ControlledGate(
-        kernel_handle::Function,
+        kernel::AbstractGate,
         control_qubits::Vector{Int},
-        kernel_qubits::Vector{Int};
-        params::Vector{<:Real}=Vector{Float64}()
         )
+
+        @assert !(kernel isa ControlledGate) "Recursive construction of ControlledGate{ControlledGate} is not allowed"
+
+        kernel_qubits=get_connected_qubits(kernel)
 
         connected_qubits=vcat(control_qubits,kernel_qubits)
 
@@ -211,18 +213,11 @@ struct ControlledGate{GateType<:AbstractGate}<:AbstractGate
             length(unique(connected_qubits)) "Control and kernel qubits must all be distinct."*
             " Received control_qubits: $control_qubits, target_qubits $kernel_qubits"
 
-        # calls function, returns Gate
-        kernel=kernel_handle(kernel_qubits..., params...) 
-        
+
         new{typeof(kernel)}(kernel,connected_qubits,control_qubits,kernel_qubits)
     end
 
-    ControlledGate(
-        kernel_handle::Function,
-        control_qubit::Int,
-        kernel_qubit::Int;
-        params::Vector{<:Real}=Vector{Float64}())=
-            ControlledGate(kernel_handle,[control_qubit],[kernel_qubit];params=params)
+    ControlledGate(kernel::AbstractGate,control_qubit::Int)=ControlledGate(kernel,[control_qubit])
 end
 
 function get_operator(gate::ControlledGate, T::Type{<:Complex}=ComplexF64) 
--- a/src/core/quantum_gate.jl
+++ b/src/core/quantum_gate.jl
@@ -12,7 +12,7 @@ Each descendant of `AbstractGate` must have at least the following fields:
 - `parameter::Real`: for parameterized gates, determines which operation is applied (e.g. rotation angles), i.e., is used in the construction of the matrix used in the application of its `Operator`.
 
 # Examples
-A struct must be defined for each new gate type, such as the following X_45 gate which
+A struct must be defined for each new `Gate` type, such as the following X_45 `Gate` which
 applies a 45° rotation about the X axis:
 
 ```jldoctest gate_struct
@@ -22,9 +22,15 @@ julia> struct X45 <: AbstractGate
 
 ```
 
-For convenience, a constructor can be defined:
+A `Gate` constructor must be defined as:
 ```jldoctest gate_struct
 julia> x_45(target::Integer) = X45(target);
+
+```
+
+along with an `Operator` constructor, with default precision `ComplexF64`, defined as:
+```jldoctest gate_struct
+julia> x_45(T::Type{<:Complex}=ComplexF64)=rotation_x(π/4, T);
 
 ```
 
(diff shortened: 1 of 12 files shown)


fSosaRey:
also, are we happy with this printout?

```
julia> ControlledGate(hadamard(),2,3)
Gate Object: ControlledGate
Connected_qubits	: [2, 3]
Operator:
(4, 4)-element Snowflake.DenseOperator:
Underlying data ComplexF64:
1.0 + 0.0im    0.0 + 0.0im    0.0 + 0.0im    0.0 + 0.0im
0.0 + 0.0im    1.0 + 0.0im    0.0 + 0.0im    0.0 + 0.0im
0.0 + 0.0im    0.0 + 0.0im    0.7071067811865475 + 0.0im    0.7071067811865475 + 0.0im
0.0 + 0.0im    0.0 + 0.0im    0.7071067811865475 + 0.0im    -0.7071067811865475 + 0.0im
```

plerouxanyon:
> also, are we happy with this printout?
> 
> ```
> julia> ControlledGate(hadamard(),2,3)
> Gate Object: ControlledGate
> Connected_qubits	: [2, 3]
> Operator:
> (4, 4)-element Snowflake.DenseOperator:
> Underlying data ComplexF64:
> 1.0 + 0.0im    0.0 + 0.0im    0.0 + 0.0im    0.0 + 0.0im
> 0.0 + 0.0im    1.0 + 0.0im    0.0 + 0.0im    0.0 + 0.0im
> 0.0 + 0.0im    0.0 + 0.0im    0.7071067811865475 + 0.0im    0.7071067811865475 + 0.0im
> 0.0 + 0.0im    0.0 + 0.0im    0.7071067811865475 + 0.0im    -0.7071067811865475 + 0.0im
> ```

Yes, I'm happy with that. The printout is a minor detail and super easy to change.

plerouxanyon:
> @plerouxanyon should we convert ControlX and ControlZ gates to ControlledGates? there would be a performance gain, as right now they use DenseOperator{4}, although only the 2x2 kernel has an effect

Yes, but let's do that in the next PR to prevent scope creep in this one.

src/core/quantum_gate.jl
```diff
@@ -101,6 +133,106 @@ get_connected_qubits(gate::AbstractGate)=[gate.target]
...
+   qubit_count: 2 
+q[1]:──*──
+       |  
+q[2]:──H──
+          
+```
+
+
+"""
+struct ControlledGate{G<:AbstractGate}<:AbstractControlledGate
```
plerouxanyon, line 178: I think we should remove `AbstractControlledGate` since we know a gate is a controlled gate if it is this type.

src/core/quantum_gate.jl
```diff
@@ -101,6 +133,106 @@ get_connected_qubits(gate::AbstractGate)=[gate.target]
...
+
+        symbol_specs=gates_display_symbols[kernel_gate_type]      
+        num_params=length(params)
+        gate_params=Dict{String,Real}() 
+        
+        for (field,param) in zip(symbol_specs[end-num_params+1:end],params)
+            gate_params[field]=param
+        end
+
+        gates_display_symbols[ControlledGate{kernel_gate_type}]=[
```
plerouxanyon, line 211: Mutating `gates_display_symbols` is hard to follow and non-obvious for someone not familiar with Snowflake.

If someone wanted to update how the `ControlledGate{XGate}` rendered, for example. They would look at `Base.show` they would, follow along and eventually see that it looks inside `gates_display_symbols`, and would be very confused to see that `ControlledGate{XGate}` is not present. They might find this if they search for all places `gates_display_symbols` is defined, but this is not obvious, and mutable changes like this tend to become fragile over time.

A good mental check to do to verify if code is readable is to put yourself in the shoes of someone that has never seen this code before. How will they navigate this code?

We need to improve this at some point, but for now, let's simply have a `get_controlled_gate_symbol` function which looks at the inner gate and adds the `*` symbols.

src/core/quantum_gate.jl
```diff
@@ -101,6 +133,106 @@ get_connected_qubits(gate::AbstractGate)=[gate.target]
...
+q[1]:──*──
+       |  
+q[2]:──H──
+          
+```
+
+
+"""
+struct ControlledGate{G<:AbstractGate}<:AbstractControlledGate
+    kernel::AbstractOperator
```
plerouxanyon, line 179: I think the kernel should be a gate (gate symbol after https://github.com/anyonlabs/Snowflake.jl/pull/232) instead of an operator.

EDIT: I understand why you made it an operator due to how the gate symbols are managed, but if we change that, then, to me, at least, it makes more sense to make it a gate.

src/core/quantum_gate.jl
```diff
@@ -92,6 +98,32 @@ q[1]:──X45──
 q[2]:───────
 ```
 
+Lastly, to enable the construction of a `ControlledGate{X45}` made with this
```
plerouxanyon, line 101: If we update the display for controlled gates, this will fall away.

src/core/quantum_gate.jl
```diff
@@ -101,6 +133,106 @@ get_connected_qubits(gate::AbstractGate)=[gate.target]
...
+                precision=precision
+            )[1]
+        ]
+        
+        new{kernel_gate_type}(kernel,connected_qubits)
+    end
+end
+
+# returns an equivalent size 4 DenseOperator built using the kernel
+function get_operator(gate::ControlledGate, T::Type{<:Complex}=ComplexF64)
```
plerouxanyon, line 226: This assumes a single-qubit single-control gate, but the definition of our gate is general. What if we have a controlled swap gate? Or a multi-controlled X gate (such as Tofolli).

src/core/quantum_gate.jl
```diff
@@ -101,6 +133,106 @@ get_connected_qubits(gate::AbstractGate)=[gate.target]
...
+# returns an equivalent size 4 DenseOperator built using the kernel
+function get_operator(gate::ControlledGate, T::Type{<:Complex}=ComplexF64) 
+    op=get_matrix(eye(4))
+    
+    op[3:4,3:4]=get_matrix(gate.kernel)
+
+    return DenseOperator(convert(Matrix{T},op))
+end
+
+get_connected_qubits(gate::ControlledGate)=gate.connected_qubits
```
plerouxanyon, line 234: The gate will only contain the target qubits, but connected qubits should also include the control qubits.

test/circuit.jl
```diff
@@ -226,3 +231,59 @@ end
...
+        ControlledGate( rotation,      connected_qubits; params=[pi,pi/2]),
+        ControlledGate( rotation_x,    connected_qubits; params=[pi/3]),
+        ControlledGate( rotation_y,    connected_qubits; params=[pi/4]),
+        ControlledGate( phase_shift,   connected_qubits; params=[pi/7]),
+        ControlledGate( universal,     connected_qubits; params=[pi/3,pi/12,pi/4]),
+    ]
+
+    circuit=QuantumCircuit(qubit_count=2,gates=gate_list)
+
+    print(circuit)
```
plerouxanyon, line 287: This isn't really a test. All that this does is make sure it doesn't crash. It doesn't test that what it does is correct. I would recommend either removing this test or extending this test to ensure that the output is what is expected.

test/transpile.jl
```diff
@@ -962,3 +962,11 @@ end
     @test !circuit_contains_gate_type(transpiled_circuit, Snowflake.Swap)
     @test simulate(circuit) ≈ simulate(transpiled_circuit)
 end
+
+@testset "CastControlledGatesToNativeTranspiler" begin
```
plerouxanyon, line 966: Native gates depend on the QPU and shouldn't be part of the transpiler name. I think renaming it to `UnsupportedGatesTranspiler` would be more apt, or any other name that does not make assumptions about the QPU.

plerouxanyon (changes requested):

src/core/quantum_gate.jl
fSosaRey: @plerouxanyon do you mean that display symbol construction will be handled in `get_display_symbol(::ControlledGate)` and not stored in `gates_display_symbols`?

src/core/quantum_gate.jl
fSosaRey: @plerouxanyon  the `connected_qubits` passed at construction of ControlledGate has both

src/core/quantum_gate.jl
plerouxanyon: Yes, since all controlled gates will be handled the same way. The controlled qubits will all be `*`, and the rest will be the regular symbols from `get_display_symbols`.

src/core/quantum_gate.jl
fSosaRey: updated to <:AbstractGate

src/core/quantum_gate.jl
fSosaRey: ok the fall-back case for arbitrairy ControlledGates is to build a DenseOperator and apply that

test/circuit.jl
fSosaRey: this was removed, each ControlledGate is tested individually for producing the correct labels

test/transpile.jl
fSosaRey: name changed

src/core/quantum_gate.jl
fSosaRey: removed

src/core/quantum_gate.jl
fSosaRey: updated

plerouxanyon (approved):

src/core/quantum_gate.jl
```diff
@@ -101,6 +126,124 @@ get_connected_qubits(gate::AbstractGate)=[gate.target]
...
+
+"""
+struct ControlledGate{GateType<:AbstractGate}<:AbstractGate
+    kernel::GateType
+    connected_qubits::Vector{Int}
+    control_qubits::Vector{Int}
+    kernel_qubits::Vector{Int}
+
+    function ControlledGate(
+        kernel_handle::Function,
```
plerouxanyon, line 202: Don't take a function that constructs a gate. Take an already constructed gate. This limits what the user can define as their own gate.

plerouxanyon (changes requested):

src/core/quantum_gate.jl
fSosaRey: ok, though this opens up the recursive ControlledGate{ControlledGate{...}}. Should I put safeguards against that?

fSosaRey: updated, with assertion error for ControlledGate{ControlledGate}

plerouxanyon (approved):

PR #248, opened 2023-06-07
Feat/ode solve abstract operator

### LINKED ISSUE #239: Make sure that all the Operators work together
There are missing functions for algebra (mul, sum, etc.) between some of the operator types.

See https://github.com/anyonlabs/Snowflake.jl/pull/237#discussion_r1215479671

fSosaRey:
Closes #239

### DIFF
--- a/docs/src/library.md
+++ b/docs/src/library.md
@@ -92,6 +92,9 @@ Bra
 DiagonalOperator
 AntiDiagonalOperator
 DenseOperator
+SwapLikeOperator
+IdentityOperator
+SparseOperator
 Base.adjoint
 is_hermitian
 Base.exp(A::AbstractOperator)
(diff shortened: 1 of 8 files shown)


plerouxanyon (changes requested):
Since this is closing https://github.com/anyonlabs/Snowflake.jl/issues/239, let's add a test to verify that all operator types work together with multiply, add, and subtract.

We can have a list of operators which are the same size (2x2) and verify that multiply, add, and subtract work as expected between them.

For example, we can test the multiply property with

```julia
get_matrix(operator_a * operator_b) == get_matrix(operator_a) * get_matrix(operator_b)
```

And replace `*` with `+` and `-` for add and subtract, respectively.

PR #259, opened 2023-06-13
Feat/basic auth

plerouxanyon:


caffeine-storm:
I'm working on a refactor right now that will let us fix the coverage issue.

Let me know if you want to land this beforehand in spite of the coverage drop.

src/anyon/qpu_interface.jl
```diff
@@ -49,17 +54,25 @@ const possible_status_list=[
     cancelled_status,
 ]
 
+function BasicAuthorization(
```
plerouxanyon, line 57: Rename to snake_case and add a verb: `encode_to_basic_authorization` or similar.

src/anyon/qpu_interface.jl
```diff
@@ -279,7 +294,11 @@ function get_status(client::Client,circuitID::String)::Status
     end
 
     if body["status"]["type"]==failed_status
-        return Status(type=body["status"]["type"],message=body["message"])
+        try
+            return Status(type=failed_status,message=body["message"])
+        catch
+            return Status(type=failed_status,message=string(body))
```
plerouxanyon, line 300: I think the message should be "no failure information available". Since this will correspond to a response of
`{"type": "failed"}`, which shouldn't happen, but it is still valid according to our OpenAPI spec.

Also, we need a test for this.

src/anyon/qpu_interface.jl
```diff
@@ -279,7 +294,11 @@ function get_status(client::Client,circuitID::String)::Status
     end
```
plerouxanyon, line 294: We should also have a test if an invalid type is returned. (Maybe we add a type in the future and the Snowflake version is old)

src/anyon/qpu_interface.jl
caffeine-storm: For debug-ability, I like including a dump of the raw response in these "should not happen" cases
:thinking: 
What if we include both? Something like `no failure information available. raw response: '${string(body)}'`?

src/anyon/qpu_interface.jl
```diff
@@ -50,65 +50,68 @@ const possible_status_list=[
...
+function decode_basic_authorization(
+    basic_auth::String
+)::Tuple{String, String}
+    parts = split(basic_auth, " ")
+    @assert length(parts) == 2
+    @assert parts[1] == "Basic"
+
+    joined_user_and_password = String(base64decode(parts[2]))
+    split_user_and_password = split(joined_user_and_password, ":")
+    @assert length(split_user_and_password) == 2
```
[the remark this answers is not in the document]
caffeine-storm: We're not working with this version anymore so I'll resolve this thread.

src/anyon/qpu_interface.jl
```diff
@@ -278,11 +293,16 @@ function get_status(client::Client,circuitID::String)::Status
...
-        return Status(type=body["status"]["type"],message=body["message"])
-    else
+    if body["status"]["type"] != failed_status
         return Status(type=body["status"]["type"])
     end
+
+    message = if haskey(body, "message")
+        body["message"]
+    else
+        "no failure information available. raw response: '$(string(body))'"
```
plerouxanyon, line 303: This is better than my suggestion. Nice.

plerouxanyon (commented):
Looks good to me!

caffeine-storm:
Feels weird to stamp it myself but @plerouxanyon opened the PR and can't self-approve ... :shrug:

PR #266, opened 2023-06-14
fix: properly reference a failed status' message

caffeine-storm:


caffeine-storm:
I found that sending a non-transpiled circuit correctly returns a message complaining about an unknown symbol _but_ our handling of that message in Snowflake had a bug.

This PR fixes the bug. Let me know if we should create an 'issue' too.

plerouxanyon (approved):

PR #267, opened 2023-06-14
feat: prefer 'shot_count' to 'num_repetitions'

caffeine-storm:


src/anyon/qpu_interface.jl
```diff
@@ -145,7 +145,7 @@ function serialize_job(circuit::QuantumCircuit,repetitions::Integer)::String
         "circuit"=>Dict{String,Any}(
                 "operations"=>Vector{Dict{String,Any}}()
             ),
-        "num_repetitions"=>repetitions
```
plerouxanyon, line 148: We should rename the variable `repetitions` to `shot_count` as well.

plerouxanyon (changes requested):

src/anyon/qpu_interface.jl
plerouxanyon: The `repetitions` parameter in `serialize_job` looks like it's still there.

plerouxanyon (changes requested):
It seems that the `repetitions` parameter in `serialize_job` was missed, but other than that, it looks good. Overall looks good

src/anyon/qpu_interface.jl
caffeine-storm: Oops... good catch! Fixed now.

alireza-najafi (approved):

alireza-najafi (approved):

plerouxanyon (approved):

Issue #274, opened 2023-06-15
cast_to_cz(::MovedGate{Swap}) is not implemented


fSosaRey:
the following test fails

```
using Snowflake
using Test

include("mock_functions.jl")

test_circuits=[
    [   
        swap(2,4),
        iswap(4,1),
        iswap_dagger(1,3),
    ]
]

@testset "AnyonYukonQPU: sequential transpilation" begin            
    qpu=AnyonYukonQPU(;host=host,user=user,access_token=access_token)
    transpiler=get_transpiler(qpu) 
    qubit_count=4
    
    for gates_list in test_circuits
        for end_pos in 1:length(gates_list)
            truncated_input=gates_list[1:end_pos]
            circuit = QuantumCircuit(qubit_count = qubit_count, gates=truncated_input)
            transpiled_circuit=transpile(transpiler,circuit)
            @test compare_circuits(circuit,transpiled_circuit)
        end
    end
end
```

PR #275, opened 2023-06-15
feat: precompile gate construction, transpilation and simulation 

fSosaRey:
block merge until #274 is fixed


plerouxanyon:
I've verified using the `--trace-compile` flag that this does indeed work. When the missing gates are added, I'm good with merging.

plerouxanyon (approved):

PR #283, opened 2023-06-22
refactor: rename snowflake to snowflurry

plerouxanyon:


plerouxanyon:
Merged as part of #284

PR #289, opened 2023-06-22
Feat/2D lattice

### LINKED ISSUE #117: Implement placing qubits onto a lattice
# Overview

Anyon's next generation of hardware, such as MonarQ, uses a qubit lattice, which we need to be able to transpile to. We need to support routing qubits on a lattice.

fSosaRey:
Closes #117


fSosaRey:
@pleroux : I marked it Draft because there is still documentation I want to add on the new features, but feedback is appreciated

fSosaRey:
rebased against the latest changes on main. It's ready for review :-)

src/anyon/anyon.jl
```diff
@@ -1,5 +1,7 @@
 using Snowflurry
 
+abstract type AbstractAnyonQPU <: AbstractQPU end
```
plerouxanyon, line 3: This is using polymorphism to share code. I would prefer if we use composition.

The polymorphism isn't inherently wrong, as both MonarQ and Yukon are AnyonQPUs. However, AnyonQPUs will continue to evolve where MonarQ and Yukon won't, as they have different evolution paths. This will cause friction to change AnyonQPU while not breaking MonardQ and Yukon. We will have a remote virtual QPU soon, which is also an AnyonQPU but won't be limited to Yukon and MonarQ's gate set.

What is more long-term is the usage of AnyonClient, but that will also only be true until we, at some point, release version 2 of our API.

Sharing code using inheritance couples the evolution of the subclasses to the base class. This might be the intent. In the classic, `Dog` and `Cat` are both `Animal`s example, I would be extremely surprised if  `Animal` evolves in such a way that makes it incompatible with `Cat` or `Dog`. Using inheritance in that example makes a lot of sense.

For example, we can define a function, `run_circuit_using_client`, that implements `run_job`, but with an `AnyonClient`, since after transpilation and validation only the client is used to send a job and wait for the result. This will allow both `AnyonYukonQPU` and `AnyonMonarQQPU` to use this method, but without coupling them to anything.

We can also factor out the CZ gateset functions and use that from both `AnyonYukonQPU` and `AnyonMonarQQPU`.

If anything is unclear, we can do a teams call tomorrow.

test/transpile.jl
```diff
@@ -370,7 +370,72 @@ end
...
+
+        ([iswap(1,2)],       1)      # no effect
+        ([iswap(2,8)],       3)
+        ([iswap(8,2)],       3)
+
+        ([control_z(5,6)],  1)      # no effect
+        ([control_z(5,10)], 5)      # target at bottom
+        ([control_z(10,1)], 9)     # target on top
+
+        ([toffoli(4,5,6)],  1)      # no effect
```
plerouxanyon, line 420: The adjacency transpilers do not have to deal with 3 qubit gates. We can remove these tests.

plerouxanyon (changes requested):

test/transpile.jl
fSosaRey: I would rather leave them in even though we wont use it with the Sequential Transpiler, as the SwapForAdjacency stage can still accomodate them. If this feature is left in it should also have a test.

src/core/connectivity.jl
```diff
@@ -0,0 +1,334 @@
...
+
+            adjacency_list[target] = neighbors
+        end
+    end
+
+    return adjacency_list
+end
+    
+#breadth-first search on 2D Lattice
+function path_search(origin::Int, target::Int, connectivity::LatticeConnectivity)
```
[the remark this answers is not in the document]
fSosaRey: by advertise, do you mean in the Docs?

fSosaRey:
I still want to add documentation now that we agree on the content

src/anyon/anyon.jl
```diff
@@ -23,28 +23,49 @@ Quantum Processing Unit:
 struct AnyonYukonQPU <: AbstractQPU
     client                  ::Client
     status_request_throttle ::Function
+    connectivity            ::AbstractConnectivity
```
plerouxanyon, line 26: The type of Yukon's Connectivity should be `LineConnectivity` as Yukon will never have any other type.

src/anyon/anyon.jl
```diff
@@ -23,28 +23,49 @@ Quantum Processing Unit:
...
 
-print_connectivity(qpu::AnyonYukonQPU,io::IO=stdout)=println(io, "1──2──3──4──5──6")
+UnionAnyonQPU=Union{AnyonYukonQPU,AnyonMonarqQPU}
 
-function Base.show(io::IO, qpu::AnyonYukonQPU)
+get_num_qubits(qpu::UnionAnyonQPU)=get_num_qubits(qpu.connectivity)
+
+get_connectivity(qpu::UnionAnyonQPU) = qpu.connectivity
+
+print_connectivity(qpu::AbstractQPU,io::IO=stdout)=print_connectivity(qpu.connectivity,Int[],io)
```
plerouxanyon, line 66: `AbstractQPU` does not have a `qpu.connectivity` member. I think this function should be implemented for `UnionAnyonQPU` and use `get_connectivity` to get the connectivity.

src/anyon/anyon.jl
```diff
@@ -219,24 +245,25 @@ Returns the transpiler associated with this QPU.
...
 ```
 """
-function get_transpiler(::AnyonYukonQPU;atol=1e-6)::Transpiler
+function get_transpiler(qpu::UnionAnyonQPU;atol=1e-6)::Transpiler
     return SequentialTranspiler([
         CastToffoliToCXGateTranspiler(),
         CastCXToCZGateTranspiler(),
         CastISwapToCZGateTranspiler(),
-        SwapQubitsForLineConnectivityTranspiler(),
+        SwapQubitsForAdjacencyTranspiler(qpu.connectivity),
```
plerouxanyon, line 258: Use `get_connectivity` here instead of assuming an implementation detail.

plerouxanyon (changes requested):
Thanks for the changes. Just a few more small things.

src/anyon/anyon.jl
fSosaRey: I changed it to `get_connectivity()` but left the type to `AbstractQPU` because a VirtualQPU also has a AllToAllConnectivity. get_connectivity is part of the AbstractQPU's interface.

fSosaRey:
@plerouxanyon Ok it should be good to go!

src/anyon/anyon.jl
```diff
@@ -23,28 +23,69 @@ Quantum Processing Unit:
...
+julia>  qpu = AnyonMonarqQPU(host="example.anyonsys.com",user="test_user",access_token="not_a_real_access_token")
+Quantum Processing Unit:
+   manufacturer:  Anyon Systems Inc.
+   generation:    MonarQ
+   serial_number: ANYK202301
+   qubit_count:   12 
+   connectivity_type:  2D-lattice
+```
+"""
+struct AnyonMonarqQPU <: AbstractQPU
```
plerouxanyon, line 61: It should be `AnyonYamaskaQPU` since `MonarQ` is a machine and not a QPU.

plerouxanyon (changes requested):

plerouxanyon (approved):

PR #300, opened 2023-08-16
Use the updated Thunderhead API to submit jobs

fSosaRey:


fSosaRey:
Marked as draft so we dont accidentally merge into main and break the user's reliance on the old API

plerouxanyon:
I had to apply the following patch to get it working: [fix-get-status.zip](https://github.com/SnowflurrySDK/Snowflurry.jl/files/12883930/fix-get-status.zip)

Also, it doesn't feel right specifying the project id in the run job, since it is specific to AnyonQPU. I think that should be moved to the AnyonQPU constructor.

plerouxanyon:
We need to ensure all tutorials, documentation and the README is updated with the project_id information

plerouxanyon (approved):

PR #302, opened 2023-09-19
Return updated `base_circuit` from `append!` & `prepend!`.

ps-pat:
This would make those methods behave like their counterparts in Base. Also, it would allow the use of the `@chain` macro.

### DIFF
--- a/src/core/qobj.jl
+++ b/src/core/qobj.jl
@@ -80,6 +80,12 @@ julia> compare_kets(ψ_0,ψ_1)
 true
 
 julia> apply_gate!(ψ_1,sigma_x(1))
+4-element Ket{ComplexF64}:
+1.5000000000000004 + 2.598076211353316im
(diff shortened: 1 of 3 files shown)


plerouxanyon (approved):

plerouxanyon:
Thank you for your contribution! It looks good to me! I'll merge when the CI/CD passes.

plerouxanyon:
It seems the documentation comments need updating before this can proceed.

ps-pat:
> It seems the documentation comments need updating before this can proceed.

Alright I'm on it.

ps-pat:
Also it seems to me that `apply_gate!` should return as well. Should I modify it as well?

plerouxanyon:
> Alright I'm on it.

Thanks!

> Also it seems to me that `apply_gate!` should return as well. Should I modify it as well?

That would be great!

plerouxanyon:
Thank you very much for the change!

PR #308, opened 2023-11-20
fix: replace old name by Snowflurry

fSosaRey:
There were a few wayward mention of "Snowflake" remaining


CODE_OF_CONDUCT.md
```diff
@@ -69,7 +69,7 @@ dispute. If you are unable to resolve the matter for any reason, or if the
 behavior is threatening or harassing, report it. We are dedicated to providing
 an environment where participants feel welcome and safe.
 
-Reports should be directed to snowflake@anyonsys.com.
+Reports should be directed to snowflurry@anyonsys.com.
```
plerouxanyon, line 72: We just need to be sure that we have this account set up (and forward mail from `snowflake@anyonsys.com` to `snowflurry@anyonsys.com`).

plerouxanyon (approved):
We can merge once we confirm that the e-mail address is set up.

fSosaRey:
I realized the secrets.`SNOWFLAKE_ID` etc were not doing anything, as the test that called them had not been updated as it was in a` try catch `block which obfuscated the error. The up-to-date `submit_circuit` is tested elsewhere. Test removed.

plerouxanyon (approved):

PR #311, opened 2023-11-22
style: use JuliaFormatter

fSosaRey:


plerouxanyon:
When I run the code through the Julia formatter, I get different results.

```julia
using JuliaFormatter
format(".")
```

Did you add any settings when formatting?

fSosaRey:
> When I run the code through the Julia formatter, I get different results.
> 
> ```julia
> using JuliaFormatter
> format(".")
> ```
> 
> Did you add any settings when formatting?

Yes, I used the VS Code extention called JuliaFormatter, but your way is better. Maybe we should add that command to the coverage.jl? WDYT?

plerouxanyon:
The formatting looks good.

However, I would add the format script in a separate script. I would be surprised if I ran a coverage script and it also formatted the code. We should also add a GitHub action to verify that the formatting is good.

We can use something like the untested script below
```bash
julia ./format.jl
git diff --exit-code
```

Alternatively, we can use the prebuilt Julia action (https://github.com/julia-actions/julia-format) to do it.

plerouxanyon (approved):

PR #314, opened 2023-12-06
feat: send qubit_count to host

fSosaRey:


plerouxanyon:
We do not need a `qubit_count` anymore if we add explicit readouts that specify the classical bits it writes to.

plerouxanyon:
> We do not need a `qubit_count` anymore if we add explicit readouts that specify the classical bits it writes to.

But we do now need to send the number of classical bits.

PR #319, opened 2023-12-13
Feat/classical bits

fSosaRey:


src/anyon/qpu_interface.jl
```diff
@@ -141,18 +141,22 @@ function serialize_job(circuit::QuantumCircuit, shot_count::Integer)::String
     for instr in get_circuit_instructions(circuit)
         if instr isa Readout
             params = Dict{String,Real}()
```
caffeine-storm, line 143: `params` doesn't seem to be used anywhere... we can remove it.
```suggestion
```

caffeine-storm (changes requested):

caffeine-storm (approved):

plerouxanyon:
I have no idea why the MacOS check is hanging. Googling around, it seems to come up that the MacOS runners sometimes hangs, but I haven't seen any fixes. We are not yet running any Snowflurry code at that point, so I am confident that it's not us.

caffeine-storm (approved):

PR #322, opened 2024-01-08
docs: include readout operations in README.md's quickstart

caffeine-storm:


README.md
```diff
@@ -103,6 +103,28 @@ The first line adds a Hadamard gate to circuit object `c` which will operate on
...
+            |
+q[2]:───────X─────────✲──
+
+```
+
+`readout(1, 1)` measures qubit 1 and assigns the result to the first bit of the resultant, classical state. `readout(2, 2)` does the same for qubit 2 and the second state bit. Snowflurry supports readout from any qubit to any result bit with a few caveats:
+- Any `readout` operation must be the final operation applied to its target qubit
+  - We plan to lift this restriction in future versions of Snowflurry
+- Separate `readout` operations must write to separate classical bits
+
```
fSosaRey, line 127: - A `QuantumCircuit` must contain at least one `readout` in order to be run on a Hardware QPU or simulated

README.md
```diff
@@ -103,6 +103,28 @@ The first line adds a Hadamard gate to circuit object `c` which will operate on
...
+# output
+Quantum Circuit Object:
+   qubit_count: 2
+q[1]:──H────*────✲───────
+            |
+q[2]:───────X─────────✲──
+
+```
+
+`readout(1, 1)` measures qubit 1 and assigns the result to the first bit of the resultant, classical state. `readout(2, 2)` does the same for qubit 2 and the second state bit. Snowflurry supports readout from any qubit to any result bit with a few caveats:
```
fSosaRey, line 123: calling it a classical state is a bit confusing. how about:
`and assigns the result to classical bit 1. `readout(2, 2)` measures the state of qubit 2 and places the result in classical bit 2.`

caffeine-storm:
This isn't the right place to introduce readouts; we're only simulating the circuits in README.md and our `simulate` doesn't support it... we should instead just add links to and call out the fact that `run_job` has extra things to consider.

PR #323, opened 2024-01-08
fix: fix asynchronous_jobs tutorial

caffeine-storm:
The tutorial now needs explicit `readout` operations.

In my testing, however, I also found that our synchronization wasn't reliable... I took a stab at improving it but we should take a close look at it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,8 +142,6 @@ using Snowflurry, SnowflurryPlots
 c = QuantumCircuit(qubit_count=2)
 push!(c, hadamard(1))
 push!(c, control_x(1, 2))
-push!(c, readout(1, 1))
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -142,6 +142,8 @@ using Snowflurry, SnowflurryPlots
 c = QuantumCircuit(qubit_count=2)
 push!(c, hadamard(1))
 push!(c, control_x(1, 2))
+push!(c, readout(1, 1))
```
fSosaRey, line 145: `simulate(c)`  does not accept Readouts. This example should not have them

fSosaRey (changes requested):

PR #326, opened 2024-01-08
docs: update README.md

caffeine-storm:
This change adds some (but not all) of the documentation for explicit readouts. It also updates/fixes a few things that make references amongst our `docs/src/tutorials` make sense.

The important bit is the explanation of `readout` and `measurement` in `docs/src/tutorials/virtual_qpu.md`. Let's make sure we're all happy with the language there as it will be users' first impression of our readouts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Pkg.add(url="https://github.com/SnowflurrySDK/SnowflurryPlots.jl", rev="main")
 
 # Getting Started
 
-The best way to learn Snowflurry is to use it! Let's try to make a two-qubit circuit which implements a [Bell/EPR state](https://en.wikipedia.org/wiki/Bell_state). We'll use Snowflurry to construct and simulate the circuit then verify our construction.
+The best way to learn Snowflurry is to use it! Let's try to make a two-qubit circuit which produces a [Bell/EPR state](https://en.wikipedia.org/wiki/Bell_state). We'll use Snowflurry to construct and simulate the circuit then verify the produced `Ket`.
 
 The quantum circuit for generating a Bell state involves a Hadamard gate on one of the qubits followed by a CNOT gate (see [here](https://en.wikipedia.org/wiki/Quantum_logic_gate) for an introduction to quantum logic gates). This circuit is shown below:
 
@@ -97,7 +97,7 @@ The first line adds a Hadamard gate to circuit object `c` which will operate on 
 
 **Note**: Unlike C++ or Python, indexing in Julia starts from "1" and not "0"!
 
-Once we've built our circuit, we can consider what, if any, transpilations we should apply. Transpilation is the process of rewriting the sequence of operations in a circuit to a new sequence. Typically, the new sequence will yield the same quantum state as the old sequence but could optimize the choice of gates used for performance or portability to different hardware. Since the circuit is relatively small and Snowflurry's simulator can handle all gates, we won't run any transpilation.
+Once we've built our circuit, we can consider if it would benefit from applying any transpilation operations. Transpilation is the process of rewriting the sequence of operations in a circuit to a new sequence. As a rule, the new sequence will yield the same quantum state as the old sequence but possibly optimizing the choice of gates used for performance, using only those gates supported by a specific hardware QPU. Since the circuit is relatively small and Snowflurry's simulator can handle all gates, we won't run any transpilation for the time being.
 
 Next, we'll simulate our circuit to see if we've built what we expect.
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -103,7 +103,7 @@ The first line adds a Hadamard gate to circuit object `c` which will operate on
 
 **Note**: Unlike C++ or Python, indexing in Julia starts from "1" and not "0"!
 
-The next step we want to take is to simulate our circuit. We do not need to transpile our circuit since our simulator can handle all gates, but for larger circuit you should consider transpilation to reduce the amount of work the simulator has to perform.
+The next step we want to take is to execute our circuit. Instead of submitting a job to a remote quantum service, we will use Snowflurry's built-in simulator. We do not need to transpile our circuit since our simulator can handle all gates, but for larger circuit you should consider transpilation to reduce the amount of work the simulator has to perform.
```
fSosaRey, line 106: I'm a little uneasy with "execute". Simulate() is the function we call for this

docs/src/tutorials/virtual_qpu.md
```diff
@@ -63,8 +63,36 @@ Quantum Circuit Object:
 q[1]:──H────*──
             |  
 q[2]:───────X──
-```               
-We can then run this circuit on the virtual qpu for let's say 101 shots. 
+```
+
+Although we've created a circuit that makes a Bell pair, we need to *meaure*
```
fSosaRey, line 68: typo: measure

fSosaRey (changes requested):
a few minor things

docs/src/tutorials/virtual_qpu.md
caffeine-storm:
```suggestion
Although we've created a circuit that makes a Bell pair, we need to *measure*
```

README.md
caffeine-storm: Yeah, `execute` isn't great in this context... I do want to be consistent with what we have above, though... [lines 50-54](https://github.com/SnowflurrySDK/Snowflurry.jl/pull/326/files/226e26c3f87199399151d510d2777dbde4568d45#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L50-L54) say "Create -> Transpile -> Execute".

Paul pointed out that "Create -> Transpile -> Execute" is _a_ typical workflow but we're walking the user through something a bit different here.... We're kinda doing "Learn Snowflurry with Create then Simulate"... I'll patch the verbiage above to reflect what we're actually working through.

README.md
```diff
@@ -103,7 +97,9 @@ The first line adds a Hadamard gate to circuit object `c` which will operate on
 
 **Note**: Unlike C++ or Python, indexing in Julia starts from "1" and not "0"!
 
-The next step we want to take is to simulate our circuit. We do not need to transpile our circuit since our simulator can handle all gates, but for larger circuit you should consider transpilation to reduce the amount of work the simulator has to perform.
+Once we've built our circuit, we can consider what, if any, transpilations we should apply. Transpilation is the process of rewriting the sequence of operations in a circuit to a new sequence. Typically, the new sequence will yield the same quantum state as the old sequence but could optimize the choice of gates used for performance or portability to different hardware. Since the circuit is relatively small and Snowflurry's simulator can handle all gates, we won't run any transpilation.
```
fSosaRey, line 100: I would change reformulate a bit:
```
Once we've built our circuit, we can consider if it would benefit from applying any transpilation operations. Transpilation is the process of rewriting the sequence of operations in a circuit to a new sequence. As a rule, the new sequence will yield the same quantum state as the old sequence but possibly optimizing the choice of gates used for performance, using only those gates supported by a specific hardware QPU. Since the circuit is relatively small and Snowflurry's simulator can handle all gates, we won't run any transpilation for the time being.
```

README.md
```diff
@@ -45,15 +45,9 @@ Pkg.add(url="https://github.com/SnowflurrySDK/SnowflurryPlots.jl", rev="main")
 
 # Getting Started
 
-A typical workflow to execute a quantum circuit on a quantum service consists of these three steps.
+The best way to learn Snowflurry is to use it! Let's try to make a two-qubit circuit which implements a [Bell/EPR state](https://en.wikipedia.org/wiki/Bell_state). We'll use Snowflurry to construct and simulate the circuit then verify our construction.
```
fSosaRey, line 48: change "implements" to "produces"

fSosaRey, line 48: change "then verify our construction" to for instance "then verify the produced Ket". Or else it seems we are verify the construction of the `circuit` object, not the resulting `Ket`.

fSosaRey (changes requested):

fSosaRey (approved):

PR #330, opened 2024-01-12
Feat/classical bits in QuantumCircuit

fSosaRey:


test/circuit.jl
```diff
@@ -2,17 +2,33 @@ using Snowflurry
...
-    @test_throws AssertionError QuantumCircuit(qubit_count = 0, instructions = [sigma_x(5)])
+    @test get_num_qubits(c) == 6
+    @test get_num_bits(c) == 2
+    @test length(get_circuit_instructions(c)) == 1
 
-    @test_throws AssertionError QuantumCircuit(qubit_count = 0)
+    @test_throws DomainError QuantumCircuit(qubit_count = 1, instructions = [sigma_x(5)])
 
+    @test_throws AssertionError(
+        "$(:QuantumCircuit) constructor requires qubit_count>0. Received: 0",
```
caffeine-storm, line 20: Are we sure we want to couple the test to the string in the error? I think it makes tests more brittle in the long run.

If we're worried about making sure we throw an error for a specific reason, we can make two circuits; a 'valid' circuit and a 'pathological' circuit; the pathological circuit can be a clone/duplicate of the valid circuit with just the invalidity added that we're testing against... WDYT?

docs/src/tutorials/basics.md
```diff
@@ -105,13 +110,12 @@ simulate(c)
 0.7071067811865475 + 0.0im
 ```
 
-The output of `simulate` function is a `Ket` object. `Ket` is a complex vector that represents the wavefunction of a quantum object such as our two-qubit system. 
+The output of `simulate` function is a `Ket` object. `Ket` is a complex vector that represents the wavefunction of a quantum object such as our two-qubit system.
```
caffeine-storm, line 113:
```suggestion
The output of the `simulate` function is a `Ket` object. A `Ket` is a complex vector that represents the wavefunction of a quantum object such as our two-qubit system.
```

caffeine-storm (commented):
I think we've got more docs to update, but here's my thoughts so far.

src/core/quantum_circuit.jl
```diff
@@ -605,6 +653,7 @@ end
 function Base.show(io::IO, circuit::QuantumCircuit, padding_width::Integer = 10)
     println(io, "Quantum Circuit Object:")
     println(io, "   qubit_count: $(get_num_qubits(circuit)) ")
+    println(io, "   bit_count: $(get_num_bits(circuit)) ")
```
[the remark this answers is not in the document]
fSosaRey: We already have an overloaded equality: `Base.isequal(QuantumCircuit, QuantumCircuit)` (which needs to include bit_count, now that I think of it)

Otherwise, printing bit_count is a matter of taste IMO. @pleroux0 ?

test/circuit.jl
fSosaRey: Hum, this is more "are we throwing the error we think we're throwing". It's happened a few times that we shadowed one error with another

src/core/quantum_circuit.jl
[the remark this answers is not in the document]
plerouxanyon: Since bit count needs to be equal for circuits to be equal, I would say that printing it is relevant.

test/circuit.jl
plerouxanyon: I think the "right" way would be to define our own custom exceptions, but not for this MR. See the last part of the following section: https://docs.julialang.org/en/v1/manual/control-flow/#Built-in-Exceptions.

fSosaRey: Ok I'll revert it to just checking the error type and add an issue for making our own custom error types

plerouxanyon: We are still checking against the error string in a few places. I'm not concerned if we are moving to custom exceptions. We can convert to a custom exception if the test string changes.

@caffeine-storm Are you happy with merging as is?

src/core/quantum_circuit.jl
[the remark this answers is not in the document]
caffeine-storm: Printing it makes sense... TIL about [the fix flag](https://documenter.juliadocs.org/stable/man/doctests/#Fixing-Outdated-Doctests) for Julia doctests so hopefully we can use that to cut down on the toil.

test/circuit.jl
```diff
@@ -38,7 +50,34 @@ end
 
     append!(c, [sigma_x(2), sigma_y(2)])
 
-    @test_throws DomainError push!(c, sigma_x(5))
+    @test_throws DomainError(
+        5,
+        "The instruction does not fit in the circuit: " * "target qubit: 5, qubit_count: 3",
```
caffeine-storm, line 55: Similar to before, checking the exact error string can become a pain later on... can we change this to check just the type (and we'll introduce/use a particular type for testing later if need be).

test/circuit.jl
```diff
@@ -38,7 +50,34 @@ end
...
+
+    @test_throws DomainError(
+        4,
+        "The instruction does not fit in the circuit: " * "target qubit: 4, qubit_count: 3",
+    ) push!(c, readout(4, 1))
+
+    @test_throws DomainError(
+        3,
+        "The instruction does not fit in the circuit: " *
+        "destination bit: 3, bit_count: 2",
```
caffeine-storm, line 79: See above; two more instances of checking error-string contents.

test/circuit.jl
caffeine-storm: Oh, nvm... we can land this change as-is and fix these incrementally.

test/circuit.jl
caffeine-storm: I'm happy to have this land.

caffeine-storm (approved):

PR #331, opened 2024-01-12
Docs/readout transpilers and functions

fSosaRey:


src/core/transpile.jl
```diff
@@ -1844,6 +1844,63 @@ end
...
+Quantum Circuit Object:
+   qubit_count: 2 
+   bit_count: 2 
+q[1]:──✲────✲──
+               
+q[2]:──────────
+               
+
+julia> transpiled_circuit = transpile(transpiler, circuit)
+ERROR: AssertionError: Found multiple Readout on qubit: 1
```
plerouxanyon, line 1900:
```suggestion
ERROR: AssertionError: Found multiple `Readouts` on qubit: 1
```

src/core/quantum_circuit.jl
```diff
@@ -267,8 +267,10 @@ yield the same output for any input, up to a global phase.
 Circuits with different ordering of gates that apply on different 
 targets can also be equivalent. 
 !!! note 
-    If there are `Readouts` are present in either `QuantumCircuit`, 
-    they are ignored by `compare_circuit`.
+    If there are `Readouts` are present on either `QuantumCircuit`,
```
plerouxanyon, line 270: Just thinking out loud, should we write `Readout`s or `Readouts`?

src/core/transpile.jl
fSosaRey: updated, along with other places we printout errors in the docs

src/core/quantum_circuit.jl
fSosaRey: I find the former looks like a typo

plerouxanyon: It does look odd, we'll then stick with `Readouts`.

plerouxanyon (approved):

PR #334, opened 2024-01-15
Feat/user agent in header

fSosaRey:
Tested that the header is sent correctly to the Thunderhead manager


src/anyon/qpu_interface.jl
```diff
@@ -61,11 +64,14 @@ function post_request(
     body::String,
 )::HTTP.Response
 
+    package_version = string(Pkg.project().version)
```
plerouxanyon, line 67: I don't think this does what we want it to do. If someone is in another project, such a `QBenchmark`, then `project().version` will return `QBenchmarks` version instead of `Snowlfurrys` version. The only reliable way to determine a package version is with `pkgversion(@__MODULE__)`, but `pkgversion` is only supported since Julia 1.9. Prior to that there are ways that rely on non-public interfaces or parsing the TOML file of the package in question. We should either drop support for Julia 1.8 and earlier, or go the TOML parsing route.

src/anyon/qpu_interface.jl
```diff
@@ -89,11 +95,14 @@ function get_request(
     access_token::String,
 )::HTTP.Response
 
+    package_version = string(Pkg.project().version)
```
plerouxanyon, line 98: See above comment

plerouxanyon (changes requested):

Project.toml
```diff
@@ -11,6 +11,7 @@ HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 JuliaFormatter = "98e50ef6-434e-11e9-1051-2b60c6c9e899"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```
plerouxanyon, line 14: When we change the `package_version` lines we should remove this

src/anyon/qpu_interface.jl
fSosaRey: updated to read from Project.toml

src/anyon/qpu_interface.jl
fSosaRey: updated

Project.toml
fSosaRey: removed, and had to add TOML

src/anyon/qpu_interface.jl
```diff
@@ -49,6 +49,12 @@ const succeeded_status = "SUCCEEDED"
 const failed_status = "FAILED"
 const cancelled_status = "CANCELLED"
 
+# Until support is dropped for Julia<1.8, calling PkgVersion(Snowflurry) 
+# is not possible. Current Snowflurry version is read from Project.toml instead
+project_toml = TOML.parsefile("Project.toml")
```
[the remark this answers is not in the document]
fSosaRey: oups, I thought I had tested it. Ok I've pushed that changed

plerouxanyon (approved):

PR #335, opened 2024-01-17
feat/inverse of operator

fSosaRey:
This allows the call to `inv(op::AbstractOperator)`, and returns the inverse of `op` as a `DenseOperator`. We could further specialize to return a `DiagonalOperator` or `AntiDiagonalOperator` if the inverse presents those features.


plerouxanyon (approved):

plerouxanyon:
Looks good to me. Since the only use-case for this is inverting the resulting operator for randomized clifford benchmarking, I think we leave it as Dense for now and implement specializations when there is a use-case for it.

PR #338, opened 2024-01-18
docs: update tutorials

fSosaRey:


docs/src/tutorials/basics.md
```diff
@@ -10,10 +10,11 @@ using Snowflurry
 
 ```
 
-We can then create an empty quantum circuit by specifying the number of qubits the circuit will involve:
+We can then create an empty [`QuantumCircuit`](@ref) by specifying the number of qubits (`qubit_count`) and classical bits (`bit_count`) the circuit will involve. 
+The classical bits (or result bits) are ordinary memory registries that store the output of a [`Readout`](@ref) operation on a particular qubit.
```
plerouxanyon, line 14: Since we are switching from plural (bits) to singular (readout), the addition of "each" makes it flow a bit better for me.

```suggestion
The classical bits (or result bits) are ordinary memory registries that each store the output of a [`Readout`](@ref) operation on a particular qubit.
```

What do you think?

plerouxanyon (approved):

PR #340, opened 2024-01-19
Feat/simulate readouts

fSosaRey:
Adds the ability to define explicit readouts and use the VirtualQPU the same way a hardware QPU would function

### DIFF
--- a/src/core/quantum_circuit.jl
+++ b/src/core/quantum_circuit.jl
@@ -1011,8 +1011,8 @@ function simulate_shots(c::QuantumCircuit, shots_count::Int = 100)
     qubit_count = get_num_qubits(c)
     c_sim = QuantumCircuit(qubit_count = qubit_count)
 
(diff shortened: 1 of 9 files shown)


src/core/quantum_circuit.jl
```diff
@@ -1017,27 +1006,151 @@ julia> simulate_shots(c, 99)
...
+                    push!(swaps_done, candidate_swap)
+                end
+            end
+        end
+
+        remapped_amplitudes[dest_state_num+1] += amplitudes[state_num+1]
+    end
+
+    return remapped_amplitudes
+end
```
plerouxanyon, line 1151: This looks a lot more complicated than it needs to be. Perhaps we could look at simplifying it?

plerouxanyon (changes requested):

plerouxanyon (approved):

PR #348, opened 2024-01-24
docs: add warning for installing from main

plerouxanyon:


fSosaRey (approved):
It is showing changes that are already merged, but I'm fine with merging as-is

PR #355, opened 2024-01-25
feat: CI with Thunderhead

alireza-najafi:
The CI.yml workflow now has a new job to run the tutorials on the Thunderhead server. This job runs on a self-hosted runner with label "thunderhead".


plerouxanyon (changes requested):
The following files still contain references to `ANYON_QUANTUM_*` variables.

* docs/src/tutorials/advanced/async_jobs.md
* docs/src/tutorials/anyon_qpu.md
* docs/src/getting_started.md

I can update them if you want.

.github/workflows/CI.yml
```diff
@@ -79,6 +79,33 @@ jobs:
         with:
           files: lcov.info
 
+  thunderhead-ci:
+    name: Thunderhead CI Test using Tutorials- ${{ github.event_name }}
+    runs-on: thunderhead
+    environment: thunderhead-dev
```
plerouxanyon, line 85: I don't think this is the intended use of `environment`. This script doesn't deploy anything, so shouldn't have an `environment`.

plerouxanyon: I've had a look, and GitHub doesn't seem to have a way to scope secrets in another way. I think we should keep this, but add a comment on why we are using the environment in a non-standard way.

fSosaRey (approved):

plerouxanyon (approved):

PR #362, opened 2024-01-30
Feat/rotation_z gate

fSosaRey:


src/core/qobj.jl
```diff
@@ -1838,5 +1838,5 @@ function compare_operators(H_0::AbstractOperator, H_1::AbstractOperator)::Bool
     δ = get_canonical_global_phase(m_0) - get_canonical_global_phase(m_1)
 
     # We need to check atol for values close to zero
-    return isapprox(m_0, exp(im * δ) .* m_1; atol = sqrt(eps(typeof(δ))))
+    return isapprox(m_0, exp(im * δ) .* m_1; atol = 1e-6)
```
fSosaRey, line 1841: I changed this because between rotation_z and it's universal equivalent there was an error 20% above `sqrt(eps())`, 1.4e-8 vs 1.2e-8.

plerouxanyon: Would this be fixed by also specifying the `rtol` instead of increasing the `atol`? According to the Julia documentation, `rtol` is zeroed instead of being `sqrt(eps(typeof(\delta)))`, which is not what I originally intended.

src/core/quantum_gate.jl
```diff
@@ -1979,6 +1979,26 @@ Base.inv(gate::RotationY) = RotationY(-gate.theta)
 
 get_gate_parameters(gate::RotationY) = Dict("theta" => gate.theta)
 
+"""
+    rotation_z(target, lambda)
+
+Return a `Gate` that applies a symmetric phase shift of `-lambda/2` and `+lambda/2` to the
```
plerouxanyon, line 1985: See discussion on #361 regarding "symmetric phase shift".

src/core/qobj.jl
fSosaRey: The following is the failing test case. I tried using rtol instead of atol but it also needs to be raised to at least `2*sqrt(eps(typeof(delta))`.


```
using Snowflurry
using Test

target = 1
theta = pi / 5

instr = rotation_z(target, theta)

universal_equivalent =
    Snowflurry.as_universal_gate(target, get_operator(get_gate_symbol(instr)))

alpha = Snowflurry.get_canonical_global_phase(get_matrix(get_operator(get_gate_symbol(instr))))

println("alpha: $alpha")

println("instr: $(exp(im*-alpha) * get_operator(get_gate_symbol(instr)))")
println("universal: $(get_operator(get_gate_symbol(universal_equivalent)))")

@test Snowflurry.compare_operators(
            get_operator(get_gate_symbol(instr)),
            get_operator(get_gate_symbol(universal_equivalent)),
        )
```

plerouxanyon: I see we are compounding errors, and then the tolerances stop making sense. For now, we should leave it as is. I've created a very-low-priority issue to allow specifying the tolerances of compare functions [here](https://github.com/SnowflurrySDK/Snowflurry.jl/issues/367).

src/core/quantum_gate.jl
fSosaRey: updated to a more standard docstring

plerouxanyon (approved):

PR #365, opened 2024-01-31
Feat/transpile controlled parametric gates

### LINKED ISSUE #358: Fix: Preserve global phase
# Overview

Snowflurry currently ignores the global phase, but the global phase matters when performing control gates.

## Tasks

- [ ] Ensure that the global phase is preserved for all operators
- [ ] Verify that `controlled(phase_shift(2,pi/2),[1])` and `controlled(z_90(2),[1])` are not equivalent.

fSosaRey:
closes #358


src/core/transpile.jl
```diff
@@ -914,6 +914,41 @@ function cast_to_rz_rx_rz(gate::Universal, target::Int)::Vector{Gate}
...
+    γ = params["theta"]
+    β = params["phi"]
+    δ = params["lambda"]
+
+    A = rotation_z(β) * rotation_y(γ / 2.0)
+
+    B = rotation_y(-γ / 2.0) * rotation_z(-(δ + β) / 2.0)
+
+    # numerical stability issues arise for rotation_z with vanishingly small phase angle
+    if abs(δ - β) > sqrt(eps(typeof(δ)))
```
plerouxanyon, line 934: I don't undestand how this gives numerical instability issues.

Given an approximate Taylor series expansion about zero of `exp(-im * x)` (The exponent of the `rotation_z`)

image

I get identical results

image

Could you add a comment on why this gives numerical instability issues, as it's not clear from these 5 lines to me?

src/core/transpile.jl
```diff
@@ -2064,3 +2101,68 @@ function transpile(
...
+            connected_qubits = get_connected_qubits(instr)
+            if length(connected_qubits) != 2
+                throw(
+                    ArgumentError(
+                        "DecomposeControlledGatesTranspiler is only implemented for single-target single-control Controlled Gates",
+                    ),
+                )
+            end
+
+            if typeof(kernel) == SigmaZ
```
plerouxanyon, line 2133: Perhaps we should just push a `control_z` gate here and let a later stage transpile that.

src/core/transpile.jl
```diff
@@ -2064,3 +2101,68 @@ function transpile(
...
+                throw(ArgumentError("use control_z() instead of Controlled(SigmaZ)"))
+            end
+
+            if typeof(kernel) == SigmaX
```
plerouxanyon, line 2137: Same as above, but with `control_x`

src/core/transpile.jl
```diff
@@ -2064,3 +2101,68 @@ function transpile(
...
+            # store existing global phase, which is cancelled by as_universal_gate()
+            a_0 = get_canonical_global_phase(get_matrix(op))
+            universal_equivalent = as_universal_gate(target, op)
+
+            (a_d, A, B, C) = decompose_universal_to_A_B_C_gates(
+                get_gate_symbol(universal_equivalent),
+                target,
+            )
+
+            α = -a_0 + a_d
```
plerouxanyon, line 2155: Let's use `\alpha` or `a` in both places for consistency.

src/core/transpile.jl
```diff
@@ -2064,3 +2101,68 @@ function transpile(
 
     return circuit
 end
+
+struct DecomposeControlledGatesTranspiler <: Transpiler end
```
plerouxanyon, line 2105: Perhaps we can rename this to `DecomposeSingleTargetSingleControlGatesTranspiler`, and then let everything pass through that is not that. Then, the `UnsupportedGatesTranspiler` can catch anything that wasn't transpiled. This way we are not as dependent on the ordering of transpilers. What do you think?

plerouxanyon (changes requested):

src/core/transpile.jl
fSosaRey: yes thats is better, updated

src/core/transpile.jl
fSosaRey: updated

src/core/transpile.jl
fSosaRey: updated

src/core/transpile.jl
fSosaRey: updated

src/core/transpile.jl
fSosaRey: I took it out, and the error i was seeing is gone

PR #371, opened 2024-02-02
Fix/2d lattice layout

### LINKED ISSUE #350: Add flexibility to Connectivity
- [ ] The qubit numbering convention in `LatticeConnectivity` does not match the numbering convention in Anyon's chips. We should change the convention to match (except for adding 1 to the index to match Julia's conventions)
- [x] An implementation of `get_adjacency_list` for `LineConnectivity` would be helpful for creating portable code

fSosaRey:
closes #350


plerouxanyon (approved):
Looks good to me!

PR #376, opened 2024-02-14
Feat/subset of lattice

fSosaRey:
in order for the transpilation to reject instructions given excluded positions on the connectivity, some refactoring was required, to add dependency injection for tests. 

`is_native_instruction` now considers the excluded positions on the connectivity

The `get_metadata` is setup so the Client is the one that retrieves it. It is hardcoded for now, but will eventually make an API call to the host server

I will update the documentation on these changes on a separate PR.

### DIFF
--- a/src/anyon/anyon.jl
+++ b/src/anyon/anyon.jl
@@ -277,11 +277,7 @@ julia> get_qubits_distance(3, 24, connectivity)
 ```
 
 """
-function get_qubits_distance(
-    target_1::Int,
-    target_2::Int,
-    c::LineConnectivity,
-)::Real
+function get_qubits_distance(target_1::Int, target_2::Int, c::LineConnectivity)::Real
     for e in c.excluded_positions
         if target_1 ≤ e ≤ target_2
             return NaN
--- a/src/anyon/anyon.jl
+++ b/src/anyon/anyon.jl
@@ -474,7 +474,7 @@ SequentialTranspiler(Transpiler[CircuitContainsAReadoutTranspiler(), ReadoutsDoN
 1──2──3──4──5──6
 ), CastSwapToCZGateTranspiler(), CompressSingleQubitGatesTranspiler(), SimplifyTrivialGatesTranspiler(1.0e-6), CastUniversalToRzRxRzTranspiler(), SimplifyRxGatesTranspiler(1.0e-6), CastRxToRzAndHalfRotationXTranspiler(), CompressRzGatesTranspiler(), SimplifyRzGatesTranspiler(1.0e-6), ReadoutsAreFinalInstructionsTranspiler(), RejectNonNativeInstructionsTranspiler(LineConnectivity{6}
 1──2──3──4──5──6
-, 6)])
+)])
 ```
 """
 function get_transpiler(
@@ -500,6 +500,6 @@ function get_transpiler(
         CompressRzGatesTranspiler(),
         SimplifyRzGatesTranspiler(atol),
         ReadoutsAreFinalInstructionsTranspiler(),
-        RejectNonNativeInstructionsTranspiler(connectivity, get_num_qubits(qpu)),
+        RejectNonNativeInstructionsTranspiler(connectivity),
     ])
 end
--- a/src/anyon/anyon.jl
+++ b/src/anyon/anyon.jl
@@ -474,8 +474,10 @@ SequentialTranspiler(Transpiler[CircuitContainsAReadoutTranspiler(), ReadoutsDoN
 )])
 ```
 """
-function get_transpiler(
-    qpu::UnionAnyonQPU;
+get_transpiler(qpu::UnionAnyonQPU; atol = 1e-6)::Transpiler =
+    get_anyon_transpiler(atol = atol, connectivity = get_connectivity(qpu))
+
+function get_anyon_transpiler(;
     atol = 1e-6,
     connectivity = get_connectivity(qpu),
 )::Transpiler
--- a/src/anyon/anyon.jl
+++ b/src/anyon/anyon.jl
@@ -187,6 +187,7 @@ end
 
 
 set_of_native_gates = [
+    Identity,
     PhaseShift,
     Pi8,
     Pi8Dagger,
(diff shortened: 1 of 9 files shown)


src/anyon/anyon.jl
```diff
@@ -32,6 +34,7 @@ struct AnyonYukonQPU <: AbstractQPU
     status_request_throttle::Function
     connectivity::LineConnectivity
     project_id::String
+    metadata::Vector{Metadata}
```
plerouxanyon, line 37: Why is metadata a vector?

src/anyon/anyon.jl
```diff
@@ -135,36 +147,26 @@ get_realm(qpu_service::UnionAnyonQPU) = get_realm(qpu_service.client)
...
-get_connectivity(qpu::UnionAnyonQPU) = qpu.connectivity
+function get_connectivity(qpu::UnionAnyonQPU)
+    md = get_metadata(qpu)
 
-print_connectivity(qpu::AbstractQPU, io::IO = stdout) =
-    print_connectivity(get_connectivity(qpu), Int[], io)
-
-function get_metadata(qpu::UnionAnyonQPU)::Dict{String,Union{String,Int}}
+    if haskey(md, "excluded_positions")
+        return with_excluded_positions(qpu.connectivity, md["excluded_positions"])
```
plerouxanyon, line 154: We need a test for this

src/anyon/anyon.jl
```diff
@@ -200,6 +203,41 @@ set_of_native_gates = [
...
+        "connectivity_type" => get_connectivity_label(qpu.connectivity),
+    )
+
+    realm = get_realm(qpu)
+    if realm != ""
+        output["realm"] = realm
+    end
+
+    if !isempty(excluded_positions)
+        output["excluded_positions"] = excluded_positions
```
plerouxanyon, line 235: We need a test for this.

src/anyon/anyon.jl
```diff
@@ -238,47 +276,72 @@ julia> get_qubits_distance(3, 24, connectivity)
 ```
 
 """
-get_qubits_distance(target_1::Int, target_2::Int, ::LineConnectivity)::Int =
+function get_qubits_distance(target_1::Int, target_2::Int, c::LineConnectivity)::Real
+    for e in c.excluded_positions
+        if target_1 ≤ e ≤ target_2
+            return NaN
```
plerouxanyon, line 282: I think we should use Infinity to indicate a disconnect.

src/anyon/anyon.jl
```diff
@@ -415,10 +471,16 @@ julia> qpu = AnyonYukonQPU(client, "project_id");
...
+), CastSwapToCZGateTranspiler(), CompressSingleQubitGatesTranspiler(), SimplifyTrivialGatesTranspiler(1.0e-6), CastUniversalToRzRxRzTranspiler(), SimplifyRxGatesTranspiler(1.0e-6), CastRxToRzAndHalfRotationXTranspiler(), CompressRzGatesTranspiler(), SimplifyRzGatesTranspiler(1.0e-6), ReadoutsAreFinalInstructionsTranspiler(), RejectNonNativeInstructionsTranspiler(LineConnectivity{6}
+1──2──3──4──5──6
+)])
 ```
 """
-function get_transpiler(qpu::UnionAnyonQPU; atol = 1e-6)::Transpiler
+function get_transpiler(
+    qpu::UnionAnyonQPU;
+    atol = 1e-6,
+    connectivity = get_connectivity(qpu),
```
plerouxanyon, line 482: This should not be part of the QPU interface. 

Rather let 'get_transpiler' dispatch on QPU type and make a 'get_anyon_transpiler' that takes the connectivity.

src/core/transpile.jl
```diff
@@ -1692,10 +1692,6 @@ function transpile(::CompressRzGatesTranspiler, circuit::QuantumCircuit)::Quantu
     )
 end
 
-struct TrivialTranspiler <: Transpiler end
```
plerouxanyon, line 1695: Why are we removing this?

test/qpu_interface.jl
```diff
@@ -975,7 +1178,12 @@ end
 
         qpu = QPU(test_client, expected_project_id, status_request_throttle = no_throttle)
 
-        histogram = transpile_and_run_job(qpu, circuit, shot_count)
+        histogram = transpile_and_run_job(
+            qpu,
+            circuit,
+            shot_count;
+            transpiler = get_transpiler(qpu; connectivity = connectivity),
```
plerouxanyon, line 1185: See above comment regarding get_transpiler. Also, this is something we can take as an argument to the constructor of AnyonQPU.

test/qpu_interface.jl
```diff
@@ -996,7 +1204,12 @@ end
             instructions = [sigma_x(qubit_count), readout(1, 1)],
         )
 
-        transpile_and_run_job(qpu, circuit, shot_count) # no error thrown
+        transpile_and_run_job(
+            qpu,
+            circuit,
+            shot_count;
+            transpiler = get_transpiler(qpu; connectivity = connectivity),
```
plerouxanyon, line 1211: See above comment

src/anyon/anyon.jl
fSosaRey: it was to get aroung the fact that struct fields are immutable, but I can also mutate an empty Dict to implement the lazy fetch. Updated

src/anyon/anyon.jl
fSosaRey: updated

src/anyon/anyon.jl
fSosaRey: this path is unreachable unless we inject excluded_positions, which seems out of place. I'll take it out for now, we can change it back when we implement the API call

src/anyon/anyon.jl
fSosaRey: updated

src/core/transpile.jl
fSosaRey: was only used as a dummy transpiler when is_native_circuit was part of `transpile_and_run_job()`. Since I moved that check to a proper transpilation stage, there was no more need for it

test/qpu_interface.jl
fSosaRey: updated to use `get_anyon_transpiler()`

src/anyon/anyon.jl
fSosaRey: added

test/qpu_interface.jl
fSosaRey: as discussed offline, having an optional argument `connectivity` when constructing QPUs conflicts with the purpose of `get_metadata()`, which will clobber anything present

plerouxanyon (approved):

PR #384, opened 2024-04-12
Feat/send git commit hash as header

fSosaRey:
I've confirmed that the hash is send if present, and nothing happens otherwise


fSosaRey:
we are closing this as the new dependencies required are too numerous for something quite peripheral to Snowflurry

PR #390, opened 2024-05-29
fix: correct off-by-one in refering to Fock basis state in Wigner funtion

### LINKED ISSUE #389: Wigner function of Fock states
There seems to be an off-by-one error when plotting Wigner functions. I've only tested with Fock states.

# Fock state plots
| | Snowflurry | QuTiP | 
| --- | --- | --- |
| n=0 | image  | image |
| n=1 | image | image |
| n=2 | image | image |

# Figure source

## Snowflurry
```julia
using Plots
using Snowflurry

N = 0
DIM = 3

density_matrix = ket2dm(fock(N, DIM))
r = LinRange(-2.5, 2.5, 100)
wig = map(r_val->wigner(density_matrix, r_val, 0.0), r)
plot(r, wig)
```

## QuTiP
```python
import matplotlib.pyplot as plt
import numpy as np
from qutip import fock, wigner

N = 0
DIM = 3

r = np.linspace(-2.5, 2.5, 100)
wig = wigner(fock(DIM, N), r, 0.0)[0]

plt.plot(r, wig)
plt.show()
```

fSosaRey:
closes #389

### DIFF
--- a/src/core/qobj.jl
+++ b/src/core/qobj.jl
@@ -1265,6 +1265,9 @@ end
 
 Returns the `i`th Fock basis of a Hilbert space with size `hspace_size` as a Ket.
 
+!!! note
+    Fock basis states numbering starts at 0.
+
 The Ket contains values of type `T`, which by default is ComplexF64.
 # Examples
 ```jldoctest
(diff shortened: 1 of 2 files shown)


jlindon-anyon (approved):
This solves the issue from #389. We should follow-up later to clarify the conventions we are using in documentation.

PR #394, opened 2024-07-03
Update gates_display_symbols for PhasedShift with corresponding arg

etiennedub:
it's a really small PR but I think I spot a mistake.
Based on the [documentation](https://snowflurrysdk.github.io/Snowflurry.jl/dev/library/quantum_gates.html#Snowflurry.phase_shift) and the [struct definition](https://github.com/SnowflurrySDK/Snowflurry.jl/blob/main/src/core/quantum_gate.jl#L2011), the arg should be "phi" and not "lambda".

### DIFF
--- a/src/core/quantum_circuit.jl
+++ b/src/core/quantum_circuit.jl
@@ -631,7 +631,7 @@ gates_display_symbols = Dict(
     RotationX => ["Rx(%s)", "theta"],
     RotationY => ["Ry(%s)", "theta"],
     RotationZ => ["Rz(%s)", "lambda"],
-    PhaseShift => ["P(%s)", "lambda"],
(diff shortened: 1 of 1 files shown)


fSosaRey:
Hi, thanks for your submission. Actually, the label "lambda" is correct, though misleading. That is the label the Thunderhead server expects for this gate type. Internally the value returned is the value set for "phi" for this particular gate. We are looking into changing the server behavior to remove this source of confusion.

fSosaRey:
In addition, the CI workflow failed as it seems github does not expose "secrets" when PRs originate from outside our organization. We are investigating workarounds.

etiennedub:
Hi, it is a bit confusing but if it's the expected behavior, you can close this PR. Is there any way I could have known this (documentation, code...)?

PR #396, opened 2024-07-08
fix: parse updated metadata response from server

fSosaRey:


plerouxanyon:
The code looks good. However, we need to deploy the updated Thunderhead manager to get the CI/CD green. Let's wait till we get that out to merge this.

plerouxanyon (approved):

PR #401, opened 2024-07-25
Feat/retry submit job

fSosaRey:


src/anyon/anyon.jl
```diff
@@ -490,23 +490,47 @@ function run_job(
     # throws error if circuit is invalid
     transpile(transpiler, circuit)
 
-    jobID =
-        submit_job(client, circuit, shot_count, get_project_id(qpu), get_machine_name(qpu))
+    return submit_with_retries(submit_and_fetch_result, client, circuit, shot_count, qpu)
```
plerouxanyon, line 493: This works, but if a network fails to get status, we only need to re-get the status, we don't have to re-submit the job as well. I think a anyon client middleware would make sense to me, but I'm happy with any solution that retries failed requests.

plerouxanyon (commented):
I've reviewed the code, but this branch contains commits from the root ZZ gate, which we need to remove.

fSosaRey:
we will adress this in a followup
https://github.com/SnowflurrySDK/Snowflurry.jl/issues/402

plerouxanyon (approved):

PR #409, opened 2024-09-03
fix: get_metadata should not error if qpu status is not "online"

fSosaRey:


plerouxanyon (commented):
Looks good, but lets also add a test that `get_metadata` succeeds with an offline QPU.

plerouxanyon (approved):

PR #410, opened 2024-09-03
fix(get_num_qubits): remove `log2` calls

jlindon-anyon:
In my testing, the calls to `log2` periodically fail on macOS Github runners. This change directly uses the julia functions `trailing_zeros` and `count_ones`, whose output does not depend on floating point comparisons.


fSosaRey (approved):

jlindon-anyon:
Moved to #411

PR #417, opened 2024-09-12
Feat/add excluded couplers to qpu metadata

npoirie1:
Adds excluded couplers to QPU metadata.

Issue #416

PR #418, opened 2024-09-13
Feat/add excluded couplers to line connectivity

npoirie1:
Adds `excluded_couplers` to `LineConnectivity`.

Issue #417

### DIFF
--- a/src/anyon/anyon.jl
+++ b/src/anyon/anyon.jl
@@ -354,20 +354,10 @@ julia> get_qubits_distance(3, 24, connectivity)
 ```
 
 """
-function get_qubits_distance(target_1::Int, target_2::Int, c::LineConnectivity)::Real
-    for e in c.excluded_positions
-        if target_1 ≤ e ≤ target_2
-            return Inf
-        end
-    end
-
-    abs(target_1 - target_2)
-end
-
 function get_qubits_distance(
     target_1::Int,
     target_2::Int,
-    connectivity::LatticeConnectivity,
+    connectivity::Union{LineConnectivity,LatticeConnectivity},
 )::Real
 
     path = path_search(target_1, target_2, connectivity)
--- a/src/anyon/anyon.jl
+++ b/src/anyon/anyon.jl
@@ -275,9 +275,9 @@ function get_metadata(client::Client, qpu::UnionAnyonQPU)::Metadata
         "status" => machineMetadata["status"],
     )
 
-    if haskey(machineMetadata, "disconnectedQubits")
+    if haskey(machineMetadata, "disconnectedCouplers")
         output["excluded_positions"] =
-            convert(Vector{Int}, machineMetadata["disconnectedQubits"])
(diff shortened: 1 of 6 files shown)


src/anyon/anyon.jl
```diff
@@ -354,20 +354,10 @@ julia> get_qubits_distance(3, 24, connectivity)
...
-    end
-
-    abs(target_1 - target_2)
-end
-
 function get_qubits_distance(
     target_1::Int,
     target_2::Int,
-    connectivity::LatticeConnectivity,
+    connectivity::Union{LineConnectivity,LatticeConnectivity},
```
fSosaRey, line 360: nice!

src/core/connectivity.jl
```diff
@@ -11,6 +11,10 @@ This connectivity type is encountered in `QPUs` such as the [`AnyonYukonQPU`](@r
 # Fields
 - `dimension         ::Int` -- Qubit count in this connectivity.
 - `excluded_positions::Vector{Int}` -- Optional: List of qubits on the connectivity which are disabled, and cannot be interacted with. Elements in Vector must be unique.
+- `excluded_connections::Vector{Tuple{Int, Int}}` -- Optional: List of couplers on the connectivity which are disabled, and cannot be interacted with. Elements in Vector must be unique.
```
fSosaRey, line 14: lets update the comment: List of connections between qubits which are disabled, over which 2-qubit gates cannot be performed.

src/core/connectivity.jl
```diff
@@ -11,6 +11,10 @@ This connectivity type is encountered in `QPUs` such as the [`AnyonYukonQPU`](@r
...
+
+Note: Every excluded connection is sorted in ascending order (i.e. connection (2, 1) will be
```
fSosaRey, line 16: the convention is: (so the rendering is uniform)
```
!!! note
    Blahblah    
```

test/qpu_interface.jl
```diff
@@ -506,24 +506,36 @@ end
 end
 
 
-@testset "AbstractConnectivity: excluded_positions" begin
+@testset "AbstractConnectivity: excluded positions and connections" begin
 
     excluded_positions = [1, 2, 3, 9, 10]
+    excluded_connections = [(2, 3), (5, 4)]
```
fSosaRey, line 512: let's use `[(2,3),(6,5)]`, so that we can also test a path between 2 valid qubits that are both not subject to an excluded position

fSosaRey (changes requested):

src/core/connectivity.jl
npoirie1: I'll change it to: "List of connections between qubits which are disabled and cannot perform 2-qubit gates."

npoirie1:
The adjacency_list has been modified so only `connectivity.excluded_positions` are not given a key.

fSosaRey (approved):

Issue #416, opened 2024-09-12
Add unavailable couplers


npoirie1:
Add functionality to prevent the use of unavailable couplers. This involves the following:

1. Adding unavailable couplers to the Anyon QPU metadata.
2. Updating the `path_search` function for the connectivity.
3. Adding a transpilation stage that throws an error if there are two-qubit gates on unavailable couplers.

PR #419, opened 2024-09-16
Feat/add excluded connections to lattice connectivity

npoirie1:
Adds `excluded_connections` to `LatticeConnectivity`.

Issue #416

### DIFF
--- a/src/core/connectivity.jl
+++ b/src/core/connectivity.jl
@@ -274,9 +274,19 @@ get_excluded_positions(c::Union{LineConnectivity,LatticeConnectivity}) =
 get_excluded_positions(connectivity::AbstractConnectivity) =
     throw(NotImplementedError(:get_excluded_positions, connectivity))
 
-get_excluded_couplers(c::LineConnectivity) =
-    c.excluded_couplers
+"""
+    get_excluded_couplers(connectivity::LineConnectivity)::Vector{Tuple{Int, Int}}
 
+Return the list of `excluded_couplers` for the `LineConnectivity`.
+"""
+get_excluded_couplers(connectivity::LineConnectivity)::Vector{Tuple{Int, Int}} =
+    connectivity.excluded_couplers
+
+"""
+    get_excluded_couplers(connectivity::AbstractConnectivity)::Vector{Tuple{Int, Int}}
+
+Throws a NotImplementedError.
+"""
 get_excluded_couplers(connectivity::AbstractConnectivity) =
     throw(NotImplementedError(:get_excluded_couplers, connectivity))
 
--- a/src/core/connectivity.jl
+++ b/src/core/connectivity.jl
@@ -17,9 +17,6 @@ This connectivity type is encountered in `QPUs` such as the [`AnyonYukonQPU`](@r
     Every excluded connection is sorted in ascending order (i.e. connection (2, 1) will be
     changed to (1, 2)).
 
-Note: Every excluded connection is sorted in ascending order (i.e. connection (2, 1) will be
-changed to (1, 2)).
-
 
 # Example
 ```jldoctest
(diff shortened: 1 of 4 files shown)


src/core/connectivity.jl
```diff
@@ -213,20 +220,84 @@ struct LatticeConnectivity <: AbstractConnectivity
...
+                ),
+            )
+        end
+
+        num_qubits = nrows * ncols
+        if sorted_connection[2] > num_qubits
+            throw(
+                AssertionError(
+                    "connection $connection must have qubits with indices less than" *
+                    " $(num_qubits + 1)",
```
fSosaRey, line 276: theres no +1 here

src/core/connectivity.jl
```diff
@@ -213,20 +220,84 @@ struct LatticeConnectivity <: AbstractConnectivity
...
+                ),
+            )
+        end
+
+        first_qubit_row_index = ((sorted_connection[1] - 1) ÷ ncols) + 1
```
fSosaRey, line 281: `mod( )` would be easier to parse than `÷` IMO

src/core/connectivity.jl
npoirie1: If there are `n` qubits, the largest qubit index is `n`. The qubit indices therefore have to be less than `n + 1`.

src/core/connectivity.jl
```diff
@@ -213,20 +220,84 @@ struct LatticeConnectivity <: AbstractConnectivity
...
+        near_index = sorted_connection[1] + ncols
+        coupler_exists =
+            sorted_connection[2] == near_index ||
+            (isodd(first_qubit_row_index) && sorted_connection[2] == near_index + 1) ||
```
fSosaRey, line 285: this logic only works for some LatticeDimensions and not others. Snowflurry.AnyonYamaskaConnectivity for instance does not follow this logic. Let's discuss how to address this offline

src/core/connectivity.jl
fSosaRey: oh! right

src/core/connectivity.jl
fSosaRey: changing the layout generation (https://github.com/SnowflurrySDK/Snowflurry.jl/pull/422) solves this problem

src/core/connectivity.jl
```diff
@@ -264,7 +338,8 @@ with_excluded_positions(
 with_excluded_positions(
     c::LatticeConnectivity,
     excluded_positions::Vector{Int},
-)::LatticeConnectivity = LatticeConnectivity(c.qubits_per_row, excluded_positions)
+)::LatticeConnectivity =
+    LatticeConnectivity(c.qubits_per_row, excluded_positions, c.excluded_connections)
```
fSosaRey, line 342: is gonna break after rebasing, needs:
`LatticeConnectivity(c.dimensions[1], c.dimensions[2], excluded_positions, c.excluded_connections)`

src/core/connectivity.jl
```diff
@@ -213,20 +220,84 @@ struct LatticeConnectivity <: AbstractConnectivity
...
     function LatticeConnectivity(
         qubits_per_row::Vector{Int},
         excluded_positions = Vector{Int}(),
+        excluded_connections = Vector{Tuple{Int,Int}}(),
     )
         nrows = maximum(qubits_per_row)
-        ncols = ceil(+(qubits_per_row...) / nrows)
+        ncols = ceil(Int, +(qubits_per_row...) / nrows)
+
+        sorted_connections =
```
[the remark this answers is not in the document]
npoirie1: Let's keep the sorted tuples.

src/core/connectivity.jl
npoirie1: The logic now works with the new layout.

fSosaRey (approved):

PR #420, opened 2024-09-16
Feat/add reject gates on excluded connections transpiler

### LINKED ISSUE #416: Add unavailable couplers

npoirie1:
Adds a transpilation stage that throws an error if a gate attempts to operate on an excluded connection.

Fixes #416


npoirie1:
I also added a test in `qpu_interface.jl` for an exception that wasn't covered in `LatticeConnecitivty`.

src/core/transpile.jl
```diff
@@ -2579,3 +2579,143 @@ function are_gates_at_excluded_positions(
...
+
+"""
+    transpile(
+        transpiler::RejectGatesOnExcludedConnectionsTranspiler,
+        circuit::QuantumCircuit
+    )::QuantumCircuit
+
+Throws a `DomainError` if an `Instruction` in the `circuit` operates on an excluded
+connection. The excluded connections are specified with the parameter `excluded_connections`
+in certain `AbstractConnectivity` objects. The `circuit` remains unchanged if no error is
```
fSosaRey, line 2595: The `circuit` remains unchanged.

fSosaRey: (you seem to imply otherwise)

src/core/transpile.jl
```diff
@@ -2579,3 +2579,143 @@ function are_gates_at_excluded_positions(
...
+    return circuit
+end
+
+function are_gates_at_excluded_connections(
+    connectivity::AllToAllConnectivity,
+    circuit::QuantumCircuit,
+)::Tuple{Bool,String}
+
+    return (false, "")
+end
```
fSosaRey, line 2693: for super simple functions I find this a bit easier to read:

```
are_gates_at_excluded_connections(
    ::AllToAllConnectivity,
    ::QuantumCircuit,
)::Tuple{Bool,String} = (false, "")
```

src/core/transpile.jl
```diff
@@ -2579,3 +2579,143 @@ function are_gates_at_excluded_positions(
...
+                message =
+                    "the $gate_name gate on qubits $qubits cannot be applied " *
+                    "since connection $connection is unavailable"
+                return (true, message)
+            end
+        end
+    end
+
+    return (false, "")
+end
```
fSosaRey, line 2721: This is a bit easier to understand, and rejects gates with more that 2 qubits (not likely to be natively supported on hardware anytime soon)

```
function are_gates_at_excluded_connections(
    connectivity::Union{LineConnectivity,LatticeConnectivity},
    circuit::QuantumCircuit,
)::Tuple{Bool,String}

    excluded_connections = get_excluded_connections(connectivity)
    for instruction in get_circuit_instructions(circuit)
        qubits = get_connected_qubits(instruction)

        if length(qubits) == 2
            candidate_connection = Tuple{Int64,Int64}(sort(qubits))
            if candidate_connection ∈ excluded_connections
                gate_name = typeof(get_gate_symbol(instruction))
                message =
                    "the $gate_name gate on qubits $qubits cannot be applied " *
                    "since connection $candidate_connection is unavailable"
                return (true, message)
            end
        end

        if length(qubits)>2
            throw(DomainError(are_gates_at_excluded_connections, "not implemented for gates with more that 2 connected qubits"))
        end

    end

    return (false, "")
end

```

src/core/transpile.jl
```diff
@@ -2579,3 +2579,143 @@ function are_gates_at_excluded_positions(
...
+    circuit::QuantumCircuit,
+)::Tuple{Bool,String}
+
+    excluded_connections = get_excluded_connections(connectivity)
+    for instruction in get_circuit_instructions(circuit)
+        qubits = get_connected_qubits(instruction)
+
+        for connection in excluded_connections
+            matching_indices = indexin(connection, qubits)
+            found_excluded_connection =
```
fSosaRey, line 2706: this works as intended, even for 3 or more connected qubits, but it is a bit misleading in our case. If there is a 3 qubit gate, say Toffoli(1,2,3), it will reject connection (1,2) and (2,3) but not (1,3), which is making assumptions on the hardware implementation. Whats worse, for the forseeable future, 3 or more qubit gates wont be supported on the hardware anyway. I suggest an easier to understand implementation:

```
function are_gates_at_excluded_connections(
    connectivity::Union{LineConnectivity,LatticeConnectivity},
    circuit::QuantumCircuit,
)::Tuple{Bool,String}

    excluded_connections = get_excluded_connections(connectivity)
    for instruction in get_circuit_instructions(circuit)
        qubits = get_connected_qubits(instruction)

        if length(qubits) == 2
            candidate_connection = Tuple{Int64,Int64}(sort(qubits))
            if candidate_connection ∈ excluded_connections
                gate_name = typeof(get_gate_symbol(instruction))
                message =
                    "the $gate_name gate on qubits $qubits cannot be applied " *
                    "since connection $candidate_connection is unavailable"
                return (true, message)
            end
        end

        if length(qubits)>2
            throw(DomainError(are_gates_at_excluded_connections, "not implemented for gates with more that 2 connected qubits"))
        end

    end

    return (false, "")
end

```

test/transpile.jl
```diff
@@ -1897,3 +1897,103 @@ end
 
     @test isequal(transpile(transpiler, valid_circuit), valid_circuit)
 end
+
+@testset "RejectGatesOnExcludedConnectionsTranspiler for all-to-all" begin
+
+    connectivity = AllToAllConnectivity()
+
+    circuit = QuantumCircuit(
```
fSosaRey, line 1905: for thoroughness, I would suggest trying some single-target gates as well

test/transpile.jl
```diff
@@ -1897,3 +1897,103 @@ end
...
+        qubit_count = 2,
+        instructions = [control_z(1, 2)],
+        name = "test-name",
+    )
+    transpiler = RejectGatesOnExcludedConnectionsTranspiler(connectivity)
+
+    @test isequal(transpile(transpiler, circuit), circuit)
+end
+
+@testset "RejectGatesOnExcludedConnectionsTranspiler for line connectivity" begin
```
fSosaRey, line 1915: add a test_throws for a Toffoli gate

test/transpile.jl
```diff
@@ -1897,3 +1897,103 @@ end
...
+    )
+
+    valid_circuit = QuantumCircuit(
+        qubit_count = 12,
+        instructions = [sigma_x(1), control_z(9, 6)],
+        name = "test-name",
+    )
+    @test isequal(transpile(transpiler, valid_circuit), valid_circuit)
+
+    valid_circuit = QuantumCircuit(
```
fSosaRey, line 1979: I dont think we should consider this valid, this implementation is tied to a hardware capability, which cannot process such a gate

src/core/transpile.jl
npoirie1: I will replace the last sentence with "The function returns the same `circuit` if no error is thrown."

fSosaRey (approved):

PR #422, opened 2024-09-17
fix: LatticeConnectivity matches hardware layout by default

fSosaRey:


src/core/connectivity.jl
```diff
@@ -117,65 +117,62 @@ This connectivity type is encountered in `QPUs` such as the [`AnyonYamaskaQPU`](
 
 # Example
 The following lattice has 4 rows, made of qubits 
-`[1, 5, 9]`, `[2, 6, 10]`, `[3, 7, 11]` and `[8, 4, 12]`, with each of those rows having 3 elements.
+`[1, 5, 9]`, `[2, 6, 10]`, `[3, 7, 11]` and `[4, 8, 12]`, with each of those rows having 3 elements.
 
-The corresponding `qubits_per_row` field is `[2, 3, 3, 3, 1]`, the number of qubits in each line
+The corresponding `qubits_per_row` field is `[1, 3, 3, 3, 2]`, the number of qubits in each line
```
npoirie1, line 122: Could we avoid having two different definitions of a row?

npoirie1 (changes requested):
Apart from the multiple definitions of a row, it looks good.

src/core/connectivity.jl
fSosaRey: renamed to qubits_per_printout_line

fSosaRey: yes you're right, thanks

npoirie1 (approved):

PR #423, opened 2024-09-20
Refactor/remove excluded positions from native transpiler

npoirie1:
Removes the check for `excluded_positions` in `RejectNonNativeInstructionsTranspiler`. The option to provide a custom list of native instructions was also added.

### DIFF
--- a/docs/src/library/qpu.md
+++ b/docs/src/library/qpu.md
@@ -18,4 +18,5 @@ get_excluded_connections
 path_search
 get_adjacency_list
 get_qubits_distance
+is_native_instruction
 ```
--- a/docs/src/library/qpu.md
+++ b/docs/src/library/qpu.md
@@ -19,4 +19,5 @@ path_search
 get_adjacency_list
 get_qubits_distance
 is_native_instruction
+is_native_circuit
 ```
--- a/src/anyon/anyon.jl
+++ b/src/anyon/anyon.jl
@@ -374,6 +374,42 @@ end
 
 const GeometricConnectivity = Union{LineConnectivity,LatticeConnectivity}
 
+"""
+    is_native_instruction(
+        gate::Union{Gate},
+        connectivity::Union{LineConnectivity,LatticeConnectivity},
+        native_gates::Vector{DataType} = set_of_native_gates,
+    )::Bool
+
+Returns `true` if the `gate` is a native instruction for the `connectivity` and the list of
+possible `native_gates`. The native gates for the Anyon QPUs are used by default.
+
+A native instruction is defined as an instruction that is in `native_gates` and that
+satisifies the `connecitivity`. It does not check to determine if the `gate` is placed at
+the `excluded_positions` or `excluded_connections` of the `connectivity`.
+
+# Example
+
+```jldoctest  
+julia> connectivity = LineConnectivity(3)
+LineConnectivity{3}
+1──2──3
+
+
+julia> is_native_instruction(control_z(1, 2), connectivity)
+true
+
+julia> is_native_instruction(control_z(1, 3), connectivity)
+false
+
+julia> is_native_instruction(toffoli(1, 2, 3), connectivity)
+false
+
+julia> is_native_instruction(toffoli(1, 2, 3), connectivity, [Snowflurry.Toffoli])
+true
+
+```
+"""
 function is_native_instruction(
     gate::Gate,
     connectivity::GeometricConnectivity,
@@ -399,7 +435,17 @@ function is_native_instruction(
     return (typeof(get_gate_symbol(gate)) in native_gates)
 end
 
+"""
+    is_native_instruction(
+        readout::Readout,
+        connectivity::Union{LineConnectivity,LatticeConnectivity},
+        native_gates::Vector{DataType} = set_of_native_gates,
+    )::Bool
 
+Always returns `true` since `readout` is a native instruction.
+
+```
+"""
 function is_native_instruction(
     readout::Readout,
     connectivity::GeometricConnectivity,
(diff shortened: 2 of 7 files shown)


src/anyon/anyon.jl
```diff
@@ -374,31 +374,179 @@ end
...
+false
+
+julia> is_native_instruction(toffoli(1, 2, 3), connectivity, [Snowflurry.Toffoli])
+true
+
+```
+"""
+function is_native_instruction(
+    gate::Gate,
+    connectivity::GeometricConnectivity,
```
fSosaRey, line 415: let's also implement `is_native_instruction` for `AllToAllConnectivity` (all gates are native)

src/anyon/anyon.jl
```diff
@@ -374,31 +374,179 @@ end
...
+   bit_count: 3 
+q[1]:──X────*──
+            |  
+q[2]:───────*──
+            |  
+q[3]:───────X──
+               
+
+julia> is_native_circuit(foreign_circuit, connectivity)
+(false, "Instruction type Gate{Snowflurry.Toffoli} with targets [1, 2, 3] is not native on the connectivity")
```
fSosaRey, line 531: let's add that the reason is it has more than 2 connected qubits, otherwise it seems odd

src/anyon/anyon.jl
```diff
@@ -374,31 +374,179 @@ end
...
+    num_qubits = get_num_qubits(connectivity)
+    return readout.connected_qubit <= num_qubits
 end
 
-is_native_instruction(readout::Readout, connectivity::GeometricConnectivity)::Bool =
-    !any(x -> x in connectivity.excluded_positions, get_connected_qubits(readout))
+"""
+    is_native_circuit(
+        circuit::QuantumCircuit,
+        connectivity::GeometricConnectivity,
```
fSosaRey, line 479: let's implement for AllToAllConnectivity

src/core/transpile.jl
```diff
@@ -2428,18 +2428,152 @@ end
 
 struct RejectNonNativeInstructionsTranspiler <: Transpiler
     connectivity::AbstractConnectivity
+    native_gates::Vector{DataType}
+
+    function RejectNonNativeInstructionsTranspiler(
+        connectivity::Union{LineConnectivity,LatticeConnectivity},
```
fSosaRey, line 2434: connectivity::GeometricConnectivity

src/anyon/anyon.jl
fSosaRey: throwing NotImplemented is fine, but let's not have it be a methodError

test/transpile.jl
```diff
@@ -1756,9 +1756,7 @@ end
 
     connectivities_and_targets = [
         (LineConnectivity(12), (1, 2))
-        (LineConnectivity(12, collect(9:12)), (1, 2))
```
fSosaRey, line 1759: I suggest leaving these in, if only to ensure the excluded positions have no effect

test/transpile.jl
```diff
@@ -1783,61 +1781,76 @@ end
...
-        LatticeConnectivity(5, 4, excluded_positions),
-    ]
+    custom_native_gates = [Snowflurry.Toffoli]
+    custom_transpiler =
+        RejectNonNativeInstructionsTranspiler(connectivity, custom_native_gates)
 
-    for connectivity in connectivities
-        for target = 1:12
-            input_gates_on_excluded_targets = [
-                phase_shift(target, pi / 3),
```
fSosaRey, line 1800: let's keep one of these to confirm there is no effect

fSosaRey (changes requested):
a few suggestions

docs/src/library/qpu.md
```diff
@@ -18,4 +18,6 @@ get_excluded_connections
 path_search
 get_adjacency_list
 get_qubits_distance
+is_native_instruction
```
fSosaRey, line 21: I'm not sure if these should be part of the public interface. We are only implementing them for some of the connectivities

src/anyon/anyon.jl
npoirie1: The function was actually returning `false` because the Toffoli gate is not in the list of native gates for the Anyon QPUs. I have changed the logic so all gates with more than 2 qubits now fail.

docs/src/library/qpu.md
npoirie1: Are you only referring to `is_native_instruction` and `is_native_circuit` or are you also referring to `path_search`, `get_adjacency_list`, and `get_qubits_distance`?

npoirie1: I would at least export `is_native_circuit` since it is useful if you're trying to make sure that the gates in your circuit won't be replaced by a larger number of gates during transpilation.

fSosaRey: yes I was referring to `is_native_instruction` and `is_native_circuit`. Ok if we add the NotImplementedErrors for AbstractConnectivity, it will be a catch-all for undefined methods

src/core/transpile.jl
```diff
@@ -2431,12 +2431,20 @@ struct RejectNonNativeInstructionsTranspiler <: Transpiler
...
-        connectivity::AbstractConnectivity,
+        connectivity::Union{LineConnectivity,LatticeConnectivity},
         native_gates::Vector{DataType} = set_of_native_gates,
     )
 
         return new(connectivity, native_gates)
     end
+
+    function RejectNonNativeInstructionsTranspiler(
+        connectivity::AllToAllConnectivity,
```
fSosaRey, line 2442: actually, it should be `connectivity::AbstractConnectivity`

fSosaRey (approved):

PR #425, opened 2024-09-23
fix: handle rounding errors for acos

npoirie1:
Rounds values `x` which are slightly greater than 1.0 or less than -1.0 back to 1.0 and -1.0, respectively, before calling `acos(x)`.


npoirie1:
I only found one call to `acos` in Snowflurry.

fSosaRey (approved):

